PR #33, opened 2020-07-07
Add NAF support for activites.

n1k0:
### Todo

- [x] add a `codes_naf` array field to the `Activite` model
- [x] attempt to infer an ERP's `Activite` from the NAF code received from INSEE's Sirene API
- [x] pre-fill the ERP form with the right `Activite` 
- [ ] fill NAF codes for all `Activite` in the admin
- [ ] export activites with NAF to default data fixtures


n1k0:
Discarding this one. NAF nomenclature is too vague and not reliable enough.

PR #47, opened 2020-09-22
Add vote evolution chart to stats page.

n1k0:


stats/views.py
```diff
@@ -43,15 +47,40 @@ def get_north_star(self):
...
 
         context["north_star"] = self.get_north_star()
         context["nb_filled_erps"] = erp_qs.count()
         context["communes"] = Commune.objects.erp_stats()[:10]
         context["nb_contributors"] = self.get_nb_contributors()
         context["top_contributors"] = self.get_top_contributors()
         context["top_voters"] = self.get_top_voters()
+        context["votes_histogram_labels"] = json.dumps(votes_data["labels"])
+        context["votes_histogram_totals"] = json.dumps(votes_data["totals"])
+        context["votes_histogram_positives"] = json.dumps(votes_data["positives"])
```
n1k0, line 84: I think there's an `escapejs` template tag we could use directly in the template.

PR #112, opened 2021-01-13
Allow users to find their drafts listed in erp search results.

n1k0:
[Trello card](https://trello.com/c/tk4gQ7gZ/439-reprise-dun-brouillon-de-mon-compte)

Allow a user to take over a draft right from the erp search results.

image


n1k0:
Please note I've also added a quickfix with the SVG logo where things weren't aligned properly.

PR #113, opened 2021-01-13
Update logo on app pages.

n1k0:
Let's use the new logo on app pages as well :)

image


n1k0:
I've just added a quickfix for a typo ([trello card](https://trello.com/c/s7S5F7Oc/469-corriger-coquille-sur-la-landing))

PR #115, opened 2021-01-13
Modern frontend

celiacheff:


n1k0 (commented):
I'm just giving a quick go at the patch, works great but I'm puzzled with the js bundle size 🤔 

This is what we have in production right now:

image

This is what I get locally with the patch:

image

I understand Parcel comes with its own code and babel polyfills are heavy, but **that** heavy? I don't think we can ship 2.2MB of js to production, we'll have to figure out how to heavily optimize this. Minification and gzipping might not be enough…

celiacheff:
Local build is not minified nor gzipped. It also include chart.js and some other js files which were included on specific pages.
Here the result for `npm run build:prod`

> static/dist/main.js.map                 ⚠️  2.84 MB     187ms
> static/dist/main.js                      865.09 KB    33.99s
> static/dist/main.css.map                 335.77 KB      78ms
> static/dist/main.css                     185.39 KB    19.20s
> static/dist/access4all.2fa98640.svg       40.35 KB     2.80s
> static/dist/access4all.9fd9699a.eot       16.18 KB     3.08s
> static/dist/access4all.b7c45ba7.ttf       16.01 KB     3.16s
> static/dist/app.753b127e.js.map           15.19 KB      59ms
> static/dist/access4all.c694b778.woff      11.88 KB     3.17s
> static/dist/app.753b127e.js                5.27 KB    11.47s
> static/dist/marker-icon.b29b8023.png       1.43 KB     3.00s
> static/dist/layers-2x.d8c4f271.png         1.23 KB     3.39s
> static/dist/layers.350ec81b.png              696 B     3.39s

Excluding map files, we have roughly 870k (before gzipping) vs 250k gzipped. There is still a lot of room for optimizations :)

bin/post_compile
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 
-python manage.py sass static/scss/style.scss static/css/style.css -t compressed
+#python manage.py sass static/scss/style.scss static/css/style.css -t compressed
```
n1k0, line 3: So this is probably the place we'll have to call npm install and build I guess? If so we'll need to check what is required to have node/npm available on the scalingo instance. We'll also need to watch how it impacts our already long deploy procedure...

run-dev.sh
```diff
@@ -18,9 +18,9 @@ trap ctrl_c INT
...
 python manage.py runserver --insecure &
 runserver_pid=$!
-python manage.py sass $SOURCE_SCSS $TARGET_CSS --watch -t compressed &
-sasswatch_pid=$!
+#python manage.py sass $SOURCE_SCSS $TARGET_CSS --watch -t compressed &
+#sasswatch_pid=$!
 
 wait ${runserver_pid} > /dev/null 2>&1
-wait ${sasswatch_pid} > /dev/null 2>&1
+#wait ${sasswatch_pid} > /dev/null 2>&1
```
n1k0, line 25: imho this whole script should just go away as we can rely on `npm start:both` to achieve the same.

static/index.js
```diff
@@ -0,0 +1,29 @@
...
+
+import 'devbridge-autocomplete';
+import 'leaflet';
+import 'leaflet.markercluster';
+import 'leaflet.locatecontrol';
+import 'leaflet-center-cross';
+import 'chart.js';
+import 'sentry';
+
+import('./js/app').then();
```
n1k0, line 14: haha why is that? :)

static/index.js
```diff
@@ -0,0 +1,29 @@
...
+import './icons/styles.css'
+import 'leaflet/dist/leaflet.css'
+import 'leaflet.markercluster/dist/MarkerCluster.Default.css'
+import 'leaflet.locatecontrol/dist/L.Control.Locate.min.css'
+
+import './vendor/selectWoo-1.0.8/css/select2.min.css'
+import './vendor/selectWoo-1.0.8/css/select2-bootstrap4.min.css'
+import './scss/style.scss';
+
+console.log('toto', $, $().autocomplete)
```
n1k0, line 25: don't forget to remove this eventually :)

static/js/jquery.js
```diff
@@ -0,0 +1,3 @@
+import jquery from 'jquery';
+
+export default (window.$ = window.jQuery = jquery);
```
n1k0, line 3: Shouldn't this be defined in `static/index.js` directly?

static/scss/style.scss
```diff
@@ -18,8 +18,7 @@ $viewport-md: 768px;
 $viewport-lg: 992px;
 $viewport-xl: 1200px;
 
-@import "../vendor/bootstrap@4.5.3/scss/bootstrap.scss";
-
+@import "~bootstrap/scss/bootstrap";
```
n1k0, line 21: For the records, this makes the build fail because we're still running the old check sass step in CI. We should probably replace this step with one with our new npm build.

templates/contrib/base.html
```diff
@@ -4,14 +4,10 @@
...
 {% block page_title %}Renseigner un établissement{% endblock%}
 
-{% block stylesheet_extras %}
-<link rel="stylesheet" href="{% static "vendor/selectWoo-1.0.8/css/select2.min.css" %}">
-<link rel="stylesheet" href="{% static "vendor/selectWoo-1.0.8/css/select2-bootstrap4.min.css" %}">
-{% endblock %}
-
 {% block javascript_extras %}
 <script src="{% static "vendor/selectWoo-1.0.8/js/select2.min.js" %}"></script>
 <script src="{% static "vendor/selectWoo-1.0.8/js/i18n/fr.js" %}"></script>
```
n1k0, line 9: Isn't there a way to import these guys in `static/index.js`?

n1k0 (commented):
This is going great, thanks for the hard work. A few comments and remarks.

bin/post_compile
celiacheff: Yes still not commited

run-dev.sh
celiacheff: Already removed but not pushed

static/scss/style.scss
celiacheff: Yes currently working on this

static/js/jquery.js
celiacheff: Well, it could work, but from what i read on the subject Parcel isn't really good with this kind of global var, and this is an recommended hack.

n1k0: Maybe just add a comment with the link to the hack/snippet ;)

n1k0:
> Local build is not minified nor gzipped. It also include chart.js and some other js files which were included on specific pages.

Ahhh, pheww. 858KB isn't great but much better than 2.3MB :D

n1k0:
Side note, this page won't load with the dev build while it does with the prod one: http://localhost:8000/app/13-la-ciotat/a/office-du-tourisme/erp/office-de-tourisme-de-la-ciotat/

I get this error in the console with dev build: `Uncaught ReferenceError: a4a is not defined`.

static/index.js
n1k0: Maybe add a short comment explaining why we want this asynchronously

templates/admin/base.html
```diff
@@ -1,5 +1,10 @@
 {% extends "admin/base.html" %}
 
+{% load static %}
+{% block extrahead %}
+    <script src="{% static "dist/main.js" %}"></script>
+{% endblock %}
```
n1k0, line 6: Remove this (see other comments)

templates/contrib/base.html
```diff
@@ -4,14 +4,10 @@
...
+<script src="{% static "dist/main.js" %}"></script>
```
n1k0, line 10: Remove this as it duplicates the import :)

templates/contrib/base.html
n1k0, line 9: These two are now imported in static/index.js so they shouldn't be included here anymore.

n1k0 (commented):
Nice! Just ensure fixing the erp details page with this call:

```js
  window.addEventListener("DOMContentLoaded", function() {
    a4a.initAppMap(
      {{ commune_json | safe }},
      {% if erp %}{{ erp.pk }}{% else %}null{% endif %},
      {{ around | default_if_none:"null" }},
      {{ geojson_list | safe }}
    );
  })
```

n1k0:
Review app deployment setup seems to work, we now face many errors with statics (see errors on sentry) :

```
ValueError/favicon.ico
Missing staticfiles manifest entry for 'dist/main.css'
```

n1k0 (approved):

PR #121, opened 2021-01-29
Add pagesjaunes API provider.

n1k0:
Trello card: https://trello.com/c/gNh2OLXx/474-investiguer-api-solocal-pagesjaunes

Work in progress.

PR #125, opened 2021-02-03
Add pagesjaunes API provider

n1k0:
Supersedes #121

image


n1k0:
@celiacheff I'd like to deploy this on staging, so people can try out the pagesjaunes API based search feature. As you can see in the diff, that makes the patch very easy to revert or disable if we don't go further. Thoughts?

n1k0:
PagesJaunes (SoLocal) won't allow their basic ERP data being stored in our database and forbid these data to be exposed as opendata anyway, so... Moving away.

PR #129, opened 2021-02-15
Modification texte cheminement exterieur

celiacheff:
[Trello](https://trello.com/c/08wLCwZx/460-modif-texte-atteindre-lentr%C3%A9e)


n1k0 (commented):
If we negate the field value, we shouldn't forget to migrate all previously entered data. Or maybe did I miss something?

celiacheff:
> If we negate the field value, we shouldn't forget to migrate all previously entered data. Or maybe did I miss something?

I wanted to negate the result in the front, instead of loosing the meaning of the field, but couldn't figure out how. Added migration instead.

erp/views.py
```diff
@@ -342,6 +342,7 @@ def get_context_data(self, **kwargs):
             )
             context["erp"] = erp
             if erp.has_accessibilite():
+                erp.accessibilite.cheminement_ext_presence = not erp.accessibilite.cheminement_ext_presence
```
n1k0, line 345: I'd rather do this in the form constructor, what do you think? Also we should definitely add a comment explaining why we're doing this.

Pipfile.lock
```diff
@@ -169,19 +161,19 @@
...
         "django-cors-headers": {
             "hashes": [
-                "sha256:5665fc1b1aabf1b678885cf6f8f8bd7da36ef0a978375e767d491b48d3055d8f",
-                "sha256:ba898dd478cd4be3a38ebc3d8729fa4d044679f8c91b2684edee41129d7e968a"
+                "sha256:1ac2b1213de75a251e2ba04448da15f99bcfcbe164288ae6b5ff929dc49b372f",
+                "sha256:96069c4aaacace786a34ee7894ff680780ec2644e4268b31181044410fecd12e"
             ],
             "index": "pypi",
-            "version": "==3.6.0"
+            "version": "==3.7.0"
```
n1k0, line 168: It's good to have dependencies updated in a single dedicated PR, could you please drop these changes on this one? Thanks!

erp/views.py
n1k0: Hold on, we shouldn't mutate the model instance but rather render its opposite value at rendering time. We need to rethink this.

n1k0:
Just had a chat with others, we're gonna postpone releasing this as we need to rethink and validate our UX approach here.

PR #142, opened 2021-03-04
Import vaccinations centers as scheduled job

celiacheff:
[Trello](https://trello.com/c/VMM9j36w/553-faire-tourner-la-commande-dimport-des-centres-de-vaccination-automatiquement-toutes-les-6h)


templates/mail/import_vaccination_notification.txt
```diff
@@ -0,0 +1,17 @@
+{% autoescape off %}
+
+Opération effectuée:
+- {{imported}} centres importés
+- {{skipped}} écartés
+
+{%if len(errors) %}
```
n1k0, line 7: Nit: please ensure consistent var and template tags formatting, they should have a leading and trailing spaces, eg. `{% foo %}` and `{{ foo }}`, not `{%foo %}` or `{{foo}}`.

erp/management/commands/import_centres_vaccination.py
```diff
@@ -29,77 +26,16 @@ def add_arguments(self, parser):
...
     def handle(self, *args, **options):  # noqa
         self.stdout.write("Importation des centres de vaccination")
 
         try:
-            if "dataset-url" in options:
-                json_data = self.retrieve_json_data(options["dataset-url"])
-            else:
-                latest_dataset_url = self.retrieve_latest_dataset_url()
-                json_data = self.retrieve_json_data(latest_dataset_url)
+            import_centres_vaccination.job(dataset_url=options["dataset-url"], verbose=options["verbose"], report=options["report"])
```
n1k0, line 39: Nit: add a trailing comma so this spreads over multiple lines for better readability.

erp/management/commands/import_centres_vaccination.py
```diff
@@ -29,77 +26,16 @@ def add_arguments(self, parser):
...
-            raise RuntimeError(
-                f"Impossible de parser les données depuis {self.api_url}:\n{err}"
-            )
-
-    def retrieve_json_data(self, dataset_url):
-        return self.get_json(dataset_url)
+        parser.add_argument(
+            "--report",
+            action="store_true",
+            help="Envoi un mail en fin d'opération",
```
n1k0, line 32: Nice 👍

erp/jobs/import_centres_vaccination.py
```diff
@@ -0,0 +1,94 @@
...
+            skipped += 1
+
+    if verbose and len(errors) > 0:
+        print("Erreurs rencontrées :")
+        for error in errors:
+            print(f"- {error}")
+
+    print("Opération effectuée:")
+    print(f"- {imported} centres importés")
+    print(f"- {skipped} écartés")
```
n1k0, line 52: How about [logging](https://docs.djangoproject.com/en/3.1/topics/logging/) instead of printing when we're in a scheduler context? Ideally we would print out when in a cli context and logging when in a scheduler one.

We're already doing that for [other jobs](https://github.com/MTES-MCT/acceslibre/blob/master/erp/jobs/check_closed_erps.py#L15).

Btw a cool feature with logging at the `error` level is that this will post to Sentry (like uncaught exceptions).

n1k0 (commented):
Looks good 👍 

A few nits and remarks

erp/jobs/import_centres_vaccination.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def job(dataset_url='', verbose=False, report=False):
+    json_data = _get_valid_data(dataset_url)
+
+    activite = Activite.objects.filter(slug="centre-de-vaccination").first()
+    if not activite:
+        raise RuntimeError("L'activité Centre de vaccination n'existe pas.")
+
+    errors, imported, skipped = _process_data(json_data, activite, verbose)
```
n1k0, line 22: As far as I can tell, `_process_data` doesn't return anything currently

erp/jobs/import_centres_vaccination.py
```diff
@@ -0,0 +1,117 @@
+"Imports vaccination centers from open data: https://www.data.gouv.fr/fr/datasets/lieux-de-vaccination-contre-la-covid-19/"
```
n1k0, line 1: Nit: shouldn't this be a comment?

erp/jobs/import_centres_vaccination.py
```diff
@@ -0,0 +1,117 @@
...
+from erp.models import Activite
+
+logger = logging.getLogger(__name__)
+
+API_URL = (
+    "https://www.data.gouv.fr/api/1/datasets/lieux-de-vaccination-contre-la-covid-19/"
+)
+
+
+def job(dataset_url="", verbose=False, is_scheduler=False):
```
n1k0, line 18: Shouldn't we better denote the first kwarg as nullable using `dataset=None`?

erp/jobs/import_centres_vaccination.py
```diff
@@ -0,0 +1,117 @@
...
+    try:
+        errors, imported, skipped = do_import(dataset_url, is_scheduler)
+
+        if is_scheduler:
+            _send_report(errors, imported, skipped)
+        else:
+            if verbose and len(errors) > 0:
+                print("Erreurs rencontrées :")
```
n1k0, line 26: How about passing an optional `logger` kwarg, which would be a function in charge of either logging or printing to the console? This is a dependency we could pass around as an argument and call it directly without using `if` statements  🤔 

And if we don't want to bother passing this arg to each function calls needing it, we should maybe consider a class.

erp/jobs/import_centres_vaccination.py
```diff
@@ -0,0 +1,117 @@
...
+        ]
+        if len(json_resources) == 0:
+            raise RuntimeError("Jeu de donnée JSON abenst.")
+        return json_resources[0]["latest"]
+    except (KeyError, IndexError, ValueError) as err:
+        raise RuntimeError(f"Impossible de parser les données depuis {API_URL}:\n{err}")
+
+
+def _retrieve_json_data(dataset_url):
+    return _get_json(dataset_url)
```
n1k0, line 82: Is this indirection necessary?

n1k0 (commented):
Here are a few more comments; most are bikeshedding, so I'm gonna approve this PR and leave it up to you to address them now or later :)

n1k0 (approved):

erp/management/commands/start_scheduler.py
```diff
@@ -16,6 +17,9 @@ def handle(self, *args, **options):
         schedule.every(notify_changed_erps.HOURS_CHECK).hours.do(
             notify_changed_erps.job, verbose=True
         )
+        schedule.every(6).hours.do(
+            import_centres_vaccination.job, verbose=True, is_scheduler=True
+        )
```
n1k0, line 22: I realize this will trigger the job every 6 hours **since the scheduler started**. What we rather need would be to run it 4 times a day at fixed times: 0h, 6h, 12h and 18h.

erp/jobs/import_centres_vaccination.py
```diff
@@ -0,0 +1,117 @@
...
+            if not is_scheduler:
+                print("S", end="", flush=True)
+            errors.append(err.__str__())
+            skipped += 1
+
+    return errors, imported, skipped
+
+
+def _send_report(errors, imported, skipped):
+    print(errors)
```
n1k0, line 106: This will print the python list of errors:

image

erp/jobs/import_centres_vaccination.py
```diff
@@ -0,0 +1,117 @@
...
+        f"[{settings.SITE_NAME}] Rapport d'importation des centres de vaccination",
+        "mail/import_vaccination_notification.txt",
+        {
+            "errors": errors,
+            "imported": imported,
+            "skipped": skipped,
+            "SITE_NAME": settings.SITE_NAME,
+            "SITE_ROOT_URL": settings.SITE_ROOT_URL,
+        },
+    )
```
n1k0, line 117: This will list errors without any information about the Erp itself; how could we at least add its name and city?

image

celiacheff: I can't get the current ERP in the catch block, I will add the data in the mapper

erp/management/commands/start_scheduler.py
celiacheff: No problem

erp/jobs/import_centres_vaccination.py
```diff
@@ -0,0 +1,122 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def outputPrintStrategy(text, end="\n", flush=False):
+    print(text, end=end, flush=flush)
+
+
+def outputVoidStrategy(text, end="\n", flush=False):
+    pass
```
n1k0, line 19: That would rather be `logger.info(text)`, and if we add a `level` arg to this interface we could even log messages at a given level (for the records, `error` will be caught by Sentry, `info` won't).

PR #144, opened 2021-03-08
Migrate in-page JavaScript code blocks to Parcel bundler

n1k0:
[Trello](https://trello.com/c/YRPlIT2U/555-modulariser-code-js-avec-parcel)


n1k0:
@celiacheff I think I went as far as I could with this already large PR. Next moves will be about componentizing maps, but that should be addressed in a dedicated PR. There will probably be other leftovers to address later as well.

I suggest you spend some time testing the patch, because we don't have func tests invovling js (yet?).

templates/common/map.html
```diff
@@ -8,13 +20,3 @@
...
-<script>
-  window.addEventListener("DOMContentLoaded", function () {
-    a4a.initAppMap(
-      {% if commune_json %}{{ commune_json | safe }}{% else %}null{% endif %},
-      {% if erp %}{{ erp.pk }}{% else %}null{% endif %},
-      {{ around | default_if_none:"null" }},
-      {{ geojson_list | safe }}
-    );
-  })
-</script>
```
n1k0, line 20: Look ma', no js in html.

PR #148, opened 2021-03-12
Migrate error monitoring to sentry.io.

n1k0:
Following the destruction ([no pun](https://www.it-connect.fr/ovhcloud-un-incendie-ravage-le-datacenter-sgb2-situe-a-strasbourg/)) of Etalab's Sentry instance, we must migrate to a new error monitoring solution. After discussing with @tristanrobert, we agreed sentry.io was a safe bet.


n1k0:
Note that we'll have to land #144 prior to refactor the way client-side error handling is implemented.

celiacheff (approved):

PR #160, opened 2021-04-13
Export OpenData to CSV / data.gouv

celiacheff:


erp/export/schemas.py
```diff
@@ -0,0 +1,61 @@
...
+    cheminement_ext_presence: bool
+    cheminement_ext_terrain_accidente: bool
+    cheminement_ext_plain_pied: bool
+    cheminement_ext_ascenseur: bool
+    cheminement_ext_nombre_marches: int
+    cheminement_ext_reperage_marches: bool
+    cheminement_ext_main_courante: bool
+    cheminement_ext_rampe: Literal["aucune", "fixe", "amovible"]
+    cheminement_ext_pente: Literal["aucune", "légère", "importante"]
+    cheminement_ext_devers: Literal["aucun", "léger", "important"]
```
n1k0, line 23: Argh, I realize we have accentuated letters in enum values 😢  Well, I suspect it's a bit late to update all this in the db…

erp/export/schemas.py
```diff
@@ -0,0 +1,61 @@
...
+    entree_pmr_informations: str
+    accueil_visibilite: bool
+    accueil_personnels: Literal["formés", "non-formés"]
+    accueil_equipements_malentendants_presence: bool
+    accueil_equipements_malentendants: Set[Literal["autres", "bim", "lsf", "scd", "lpc"]]
+    accueil_cheminement_plain_pied: bool
+    accueil_cheminement_ascenseur: bool
+    accueil_cheminement_nombre_marches: int
+    accueil_cheminement_reperage_marches: bool
+    accueil_cheminement_main_courante: bool
```
n1k0, line 50: I'm surprised all bool fields are not optional, while they are in our db model ; shouldn't these be Optional[bool]? If not, would that mean that all bool fields are required in the schema?

n1k0 (commented):
Promising work 👍  A few preliminary questions, remarks and comments.

erp/export/schemas.py
```diff
@@ -0,0 +1,61 @@
+from dataclasses import dataclass
+from typing import Literal, Set
+
+
+@dataclass
+class OfficialSchema:
```
n1k0, line 6: I'm also realizing we are duplicating the schema definition while we have it defined as JSON in https://github.com/MTES-MCT/acceslibre-schema… Wouldn't it make sens to all this while parsing the `schema.json`? Redefining it programmatically here may introduce a risk of desynchronization between the two definitions.

erp/jobs/export_to_datagouv.py
```diff
@@ -0,0 +1,10 @@
+from erp.export.export import export_to_csv
+from erp.export.mappers import map_erp_to_json_schema
+from erp.models import Erp
+
+
+def job(*args, **kwargs):
+    with open('export.csv', 'w', newline='') as csv_file:
+        erps = Erp.objects.select_related('accessibilite').filter(published=True)
```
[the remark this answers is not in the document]
celiacheff: Right. It's actually handled after, but having the filter at the beginning id more logical.

erp/models.py
```diff
@@ -945,7 +945,7 @@ class Meta:
         choices=schema.NULLABLE_BOOLEAN_CHOICES,
         verbose_name="Entrée de plain-pied",
     )
-    # Nombre de marches
+    # Nombre de marchesschema.PENTE_LENGTH_CHOICES
```
n1k0, line 948: I suspect this is a typo

erp/export/base-schema.json
```diff
@@ -0,0 +1,54 @@
...
+      "title": "Etalab Licence Ouverte 2.0",
+      "name": "etalab-2.0",
+      "path": "https://spdx.org/licenses/etalab-2.0.html"
+    }
+  ],
+  "resources": [
+    {
+      "title": "Fichier valide (CSV)",
+      "name": "exemple-valide-csv",
+      "path": "https://github.com/MTES-MCT/acceslibre-schema/raw/v0.0.1/exemple-valide.csv"
```
n1k0, line 25: We should ensure creating a card to not forget doing this

erp/export/base-schema.json
```diff
@@ -0,0 +1,54 @@
...
+    "a11y"
+  ],
+  "countryCode": "FR",
+  "homepage": "https://github.com/MTES-MCT/acceslibre-schema",
+  "path": "https://github.com/MTES-MCT/acceslibre-schema/raw/v0.0.1/schema.json",
+  "licenses": [
+    {
+      "title": "Etalab Licence Ouverte 2.0",
+      "name": "etalab-2.0",
+      "path": "https://spdx.org/licenses/etalab-2.0.html"
```
n1k0, line 18: This is still an open question, though I suspect we should only go with the most permissive license. Let's just not forget to obtain a decision made by the product team.

erp/export/base-schema.json
```diff
@@ -0,0 +1,54 @@
...
+  "version": "0.0.1",
+  "contributors": [
+    {
+      "title": "Caleb Eliacheff",
+      "email": "caleb.eliacheff@beta.gouv.fr",
+      "organisation": "MTES-MCT",
+      "role": "author"
+    },
+    {
+      "title": "Nicolas Pierrault",
```
n1k0, line 40: It's `Perriault` ;)

erp/export/create_schema.py
```diff
@@ -0,0 +1,93 @@
+from frictionless import Schema, Field
+
+from erp import schema
+from erp.export.models import ETALAB_SCHEMA_FIELDS
+from erp.schema import FIELDS
+
+
+def generate_schema(base="base-schema.json", outfile="schema.json"):
+    table_schema = Schema(base)
+    # table_schema["version"] = "0.0.1"
```
n1k0, line 10: This raises a good question; how and where should we manage the schema version? Ideally in a single place…

erp/export/base-schema.json
```diff
@@ -0,0 +1,54 @@
...
+      "role": "author"
+    }
+  ],
+  "fields": [
+    {
+      "name":"id",
+      "description":"Identifiant unique",
+      "type":"string",
+      "example":""
+    }
```
n1k0, line 52: Maybe we shouldn't have this by default in the base template?

erp/export/create_schema.py
```diff
@@ -0,0 +1,93 @@
...
+
+    table_schema.to_json(outfile)
+
+
+def create_field(field_name, field):
+    constraints = get_constraints(field_name, field)
+    field = Field(
+        name=field_name,
+        type=map_types(field.get("type")),
+        description=field.get("help_text_ui") or field.get("description"),
```
n1k0, line 27: Side note, we should ensure always having a non-null `help_text_ui` here; how about raising an exception if it's missing? I like the idea of a self-validating schema generator

erp/export/create_schema.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def get_linked_enum(field_name):
+    if field_name == "cheminement_ext_devers":
+        return schema.DEVERS_CHOICES
+    if field_name == "entree_marches_rampe":
+        return schema.RAMPE_CHOICES
+    if field_name == "entree_dispositif_appel_type":
+        return schema.DISPOSITIFS_APPEL_CHOICES
+    if field_name == "accueil_personnels":
```
n1k0, line 103: How about exposing field choices to the `erp.schema.FIELDS` dict entries?

erp/export/export.py
```diff
@@ -0,0 +1,11 @@
+import csv
+from dataclasses import asdict
+from typing import List
+
+
+def export_to_csv(file, headers, mapped_data: List):
+    csv_writer = csv.DictWriter(file, fieldnames=headers)
```
n1k0, line 7: We should probably double-check the defaults (separator, quoting) used by DictWriter to ensure it matches Etalab's expectations.

erp/export/mappers.py
```diff
@@ -0,0 +1,115 @@
...
+    headers = ETALAB_SCHEMA_FIELDS
+
+    results = []
+    for erp in erps:
+        try:
+            erp.accessibilite
+        except Accessibilite.DoesNotExist:
+            continue
+
+        o = EtalabModel(
```
n1k0, line 35: Possibily silly idea: how about exposing a `to_etalab_json` method in the `Erp` model? so we could write:

```python
results = [erp.to_etalab_json() for erp in erps]
```

erp/export/models.py
```diff
@@ -0,0 +1,72 @@
...
+    entree_dispositif_appel: bool = None
+    entree_dispositif_appel_type: Set[
+        Literal["bouton", "sonnette", "interphone", "visiophone"]
+    ] = None
+    entree_balise_sonore: bool = None
+    entree_aide_humaine: bool = None
+    entree_largeur_mini: int = None
+    entree_pmr: bool = None
+    entree_pmr_informations: str = None
+    accueil_visibilite: bool = None
```
n1k0, line 44: I'm totally foreign to static typing in Python, but I'm surprised `bool` typed members are nullable by design, are you sure we don't need an `Option`, `Maybe` or equivalent here?

n1k0 (commented):
A few more comments and remarks

erp/export/models.py
celiacheff: I want to force the explicit declaration of values, which can be None.

erp/export/mappers.py
celiacheff: Not sure about, this, it would be kind of a leak of infra in the model. I'd rather return a a predefined dict. Or maybe be more OOP and inject a presenter.

erp/export/models.py
n1k0: So it's not a `bool`, but an `Optional[bool]`, right?

tests/erp/export/test_export_integration.py
```diff
@@ -0,0 +1,38 @@
...
+    finally:
+        os.remove(csv_file.name)
+
+
+@pytest.mark.django_db
+def test_check_modified_schema(db):
+    # use stringio to test schema generation, then compare
+    # maybe_new_schema = generate_schema()
+    # assert maybe_new_schema == actual_schema
+    ...
```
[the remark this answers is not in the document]
n1k0: TIL!

tests/erp/export/test_export_integration.py
```diff
@@ -0,0 +1,38 @@
...
+
+@pytest.mark.django_db
+def test_csv_creation(db):
+    try:
+        with open("export-test.csv", "w", newline="") as csv_file:
+            erps = Erp.objects.having_a11y_data().all()[0:10]
+            headers, mapped_data = map_erp_to_json_schema(erps)
+            export_to_csv(csv_file, headers, mapped_data)
+
+            assert Path("export-test.csv").exists() is True
```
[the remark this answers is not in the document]
celiacheff: What do you mean ?

erp/export/export.py
```diff
@@ -0,0 +1,21 @@
...
+    csv_writer.writeheader()
+
+    for erp in mapped_data:
+        data = asdict(erp)
+        # TODO: better to be in dict_factory
+        # Solve the quoting problem: frictionless only validate with "[""value""]",
```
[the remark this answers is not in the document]
n1k0: I highly doubt the JSON string escaping of array values is natively handled by Excel, as Excel doesn't natively deal with arrays for cell values. It's not because Python's DictWriter claims they support an "Excel" setting that it will magically implement and support this kind of frictionless/tableschema specificity… At least we should definitely test what's the contributor experience is when crafting a file in Excel and trying to import it before publishing the schema.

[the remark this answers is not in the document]
celiacheff: I'm just using JSON to get the correct ouput (double quotes surrounding and double quotes escaped/doubled inside). I've just checked with other schemas on ETALAB, and they also do like this.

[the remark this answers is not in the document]
n1k0: Haha OK if Etalab is okay with it we should probably go with it 👍

tests/erp/export/test_export_unit.py
```diff
@@ -0,0 +1,73 @@
...
+    export_schema_to_csv(file, example_data, EtalabMapper)
+    file.seek(0)
+    reader = csv.DictReader(file, fieldnames=first_row)
+    next(reader)  # Skip headers
+
+    erp_0 = next(reader)
+    assert erp_0["transport_station_presence"] == str(
+        mapped_data[0].transport_station_presence
+    )
+    # assert erp_0["commentaire"] == str(mapped_data[0].commentaire)
```
n1k0, line 44: Why is this commented?

tests/erp/export/test_export_unit.py
```diff
@@ -0,0 +1,73 @@
...
+    erp_1 = next(reader)
+    assert erp_1["transport_station_presence"] == str(
+        mapped_data[1].transport_station_presence
+    )
+    # assert erp_1["commentaire"] == str(mapped_data[1].commentaire)
```
n1k0, line 49: Why is this commented?

tests/erp/export/test_export_unit.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # Validate data with schema
```
n1k0, line 51: Can this patch land without this being implemented? What's missing? If we don't plan on landing this patch with this part, could we remove the comment? Or reformat the comment expliciting that this needs implementation, and why?

tests/erp/export/test_export_unit.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+# Test/check invalid data ex: no accessibility object -> should be ignored
+# Test transformed/mapped data
+
+# Test upload data.gouv
```
n1k0, line 57: Same remark

tests/erp/export/test_export_integration.py
```diff
@@ -0,0 +1,37 @@
...
+        assert result.get("errors") == []
+    finally:
+        os.remove(dest_path)
+
+
+@pytest.mark.django_db
+def test_check_modified_schema(db):
+    # use stringio to test schema generation, then compare
+    # maybe_new_schema = generate_schema()
+    # assert maybe_new_schema == actual_schema
```
n1k0, line 36: Why isn't this implemented? What's missing for implementation? Can we land this patch without this test, or should we keep working on it?

erp/export/utils.py
```diff
@@ -0,0 +1,50 @@
...
+
+from erp.models import Erp
+
+
+@dataclass(frozen=True)
+class BaseExportMapper:
+    @staticmethod
+    @abstractmethod
+    def headers():
+        ...
```
n1k0, line 13: Nit: Why are we using Python Ellipsis here and elsewhere? Doesn't seem like a valid usage… https://docs.python.org/dev/library/constants.html#Ellipsis 

I'd rather use a standard `pass` statement

n1k0 (approved):
I think we should land this patch, it's been in the work for a long time now. I've added some remarks to clean it up, which should be addressed before landing.

PR #163, opened 2021-04-14
Migrate fontastic icons to icomoon.

n1k0:
[Fontastic](https://fontastic.me), which powers our iconsets, looks closed. This PR migrates our icons to use [Icomoon](https://icomoon.io/), an equivalent free service providing self-hostable icon resources.

I could find equivalent or alternative icons to the fontastic sets we were using, so this should be quite safe. But better spend some time checking everything looks okay locally before pushing to staging.


celiacheff (approved):
Maybe we could remove demo files: readme, demo.css, demo.js, demo.html ? Otherwise, new icons looks nice.

n1k0:
> Maybe we could remove demo files: readme, demo.css, demo.js, demo.html ?

We already exposed an html page listing fontastic icons: https://github.com/MTES-MCT/acceslibre/tree/3c4fc76ab9fbcaf92ffd4a4a2a8592da33009d9f/static/icons

This has always been proven useful when we wanted to access or share the list of all available icons, I think we should keep it. Also there's no sensitive information of any sort exposed in these files, so I can't see a risk of keeping it as-is.

PR #167, opened 2021-04-17
Opt out app of Google FLoC.

n1k0:
see https://paramdeo.com//blog/opting-your-website-out-of-googles-floc-network


celiacheff (approved):
Insane. I wonder if we could ask our cloud provider to add an option to activate this header at the proxy level.

PR #172, opened 2021-04-26
Improve slope questions 2

celiacheff:


n1k0 (approved):
Looks good and works nicely locally. Don't forget to reset the staging database with a compatible db schema state before landing this patch onto master and deploying it to staging.

PR #176, opened 2021-05-05
Publication directe à la fin de la saisie d'une fiche

celiacheff:
[Trello](https://trello.com/c/S24NO5Bb/510-publication-directe-%C3%A0-la-fin-de-la-saisie-dune-fiche)

### DIFF
--- a/erp/forms.py
+++ b/erp/forms.py
@@ -607,6 +607,8 @@ class PublicPublicationForm(forms.ModelForm):
     published = forms.BooleanField(
         label="Je souhaite mettre en ligne cette fiche d'établissement immédiatement",
         required=False,
+        initial=True,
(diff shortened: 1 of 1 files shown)


erp/forms.py
```diff
@@ -607,6 +607,8 @@ class Meta:
     published = forms.BooleanField(
         label="Je souhaite mettre en ligne cette fiche d'établissement immédiatement",
         required=False,
+        initial=True,
+        widget=forms.CheckboxInput(attrs={"checked": True}),
```
n1k0, line 611: Why is this necessary? Also, are you sure it's not `"checked": "checked"`, as in HTML?

celiacheff: Same result. Inital=True is not sufficient, and doesn't check the box, even if it could.

n1k0 (approved):

PR #182, opened 2021-05-27
Importer les brigades de gendarmerie

celiacheff:
[Trello](https://trello.com/c/hzzSK0zN/610-importer-les-brigades-de-gendarmerie)

Test command `./manage.py import_gendarmeries --verbose`


erp/models.py
```diff
@@ -261,6 +261,7 @@ class Erp(models.Model):
     SOURCE_SIRENE = "sirene"
     SOURCE_TH = "tourisme-handicap"
     SOURCE_VACCINATION = "centres-vaccination"
+    SOURCE_GENDARMERIE = "gendarmerie"
```
n1k0, line 264: Nit: We should keep the constants ordered alphabetically

erp/models.py
```diff
@@ -395,6 +397,13 @@ class Meta:
         verbose_name="Courriel",
         help_text="Adresse email permettant de contacter l'ERP",
     )
+    private_contact_email = models.EmailField(
+        max_length=255,
+        null=True,
+        blank=True,
+        verbose_name="Courriel privé",
+        help_text="Email permettant de notifier le fournisseur en cas de modifications par un tier",
+    )
```
n1k0, line 406: As discussed this morning, we don't need this new field as we can already use the email address of the user account associated to the ERP entry for notifications.

erp/import_datasets/base_mapper.py
```diff
@@ -0,0 +1,24 @@
...
+    def dataset_url(self) -> str:
+        pass
+
+    @property
+    @abc.abstractmethod
+    def activite(self) -> str:
+        pass
+
+    @abc.abstractmethod
+    def fetch_data(self) -> List[Any]:
```
n1k0, line 19: I'm puzzled with the use of `Any` here; shouldn't we return a more specific type instead?

erp/import_datasets/fetcher_strategy.py
```diff
@@ -0,0 +1,64 @@
...
+    @abstractmethod
+    def fetch(self, uri) -> Any:
+        pass
+
+
+class JsonFetcher(Fetcher):
+    def fetch(self, url):
+        try:
+            return requests.get(url).json()
+        except requests.exceptions.RequestException as err:
```
n1k0, line 21: We need to catch [json.JSONDecodeError](https://docs.python.org/3/library/json.html#json.JSONDecodeError) exceptions as well.

erp/jobs/import_gendarmerie.py
```diff
@@ -0,0 +1,47 @@
...
+            ).job(verbose=verbose)
+            if self.mail_notification:
+                self._send_report(imported, skipped, errors)
+        except RuntimeError as err:
+            fatal(err)
+
+    def _send_report(self, imported, skipped, errors):
+        mailer.mail_admins(
+            f"[{settings.SITE_NAME}] Rapport d'importation des gendarmeries",
+            "mail/import_vaccination_notification.txt",
```
n1k0, line 39: For gendarmeries, we'd expect a dedicated mail templates instead of using the one for vaccination centers. Maybe create a more generic template?

erp/mapper/gendarmerie.py
```diff
@@ -0,0 +1,140 @@
...
+            self.erp.telephone = record["telephone"]
+            self.erp.code_insee = record["code_commune_insee"]
+            self.erp.code_postal = record["code_postal"]
+            self.erp.numero = numero
+            self.erp.voie = voie
+            self.erp.geom = self._import_coordinates(record)
+            self.erp.site_internet = record["url"]
+            self.erp.nom = record["service"]
+        except (KeyError, IndexError) as err:
+            raise RuntimeError("Impossible d'extraire des données: " + str(err))
```
n1k0, line 101: Nit: We could use f-strings here (`f"{x}"`)

erp/mapper/vaccination.py
```diff
@@ -92,18 +103,43 @@ def process(self, activite):
         try:
             # Save erp instance
             self.erp.published = True
-            self.erp.save()
+            # self.erp.save()
```
n1k0, line 106: This comment should now be removed.

erp/mapper/vaccination.py
```diff
@@ -92,18 +103,43 @@ def process(self, activite):
...
 
             # Attach an Accessibilite to newly created Erps
             if not self.erp.has_accessibilite():
                 accessibilite = Accessibilite(erp=self.erp)
                 accessibilite.commentaire = commentaire
-                accessibilite.save()
+                # accessibilite.save()
```
n1k0, line 112: This comment should now be removed.

erp/mapper/vaccination.py
```diff
@@ -231,6 +267,7 @@ def _retrieve_commune_ext(self):
             )
 
         if not commune_ext:
+            print(self.erp.code_insee, self.erp.code_postal)
```
n1k0, line 270: Do we want to keep this print statement?

tests/erp/mapper/test_gendarmerie.py
```diff
@@ -0,0 +1,63 @@
...
+    yield _factory
+
+
+def test_basic_stats(import_dataset, gendarmeries_valid):
+    imported, skipped, errors = import_dataset().job()
+    erp = Erp.objects.filter(
+        source_id=gendarmeries_valid[0]["identifiant_public_unite"]
+    ).first()
+    assert erp is not None
+    assert imported == 3, skipped == 0
```
n1k0, line 31: You can't chain multiple asserts that way:

```py
In [1]: assert True is True

In [2]: assert True is True, False is True

In [3]: assert False is True
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-3-826790e2fb9e> in <module>
----> 1 assert False is True
```

n1k0 (commented):
Promising patch! This is a first preliminary review, will test out the patch later but some remarks can be already addressed meanwhile.

erp/mapper/vaccination.py
celiacheff: I had a bug on communes, which i fact was due to some INSEE codes we don't have in DB (some exotic ones). Forgot to remove this.

tests/erp/mapper/test_gendarmerie.py
celiacheff: Wow thanks, I tried to refactor my tests and didn't noticed that. This was too reckless.

erp/mapper/gendarmerie.py
```diff
@@ -0,0 +1,140 @@
...
+            commune_ext = Commune.objects.filter(
+                code_postaux__contains=[self.erp.code_postal]
+            ).first()
+        else:
+            raise RuntimeError(
+                f"Champ code_insee et code_postal nuls (commune: {self.erp.commune})"
+            )
+
+        if not commune_ext:
+            raise RuntimeError("Impossible de résoudre la commune: ")
```
n1k0, line 123: We're missing the commune here

erp/import_datasets/base_mapper.py
celiacheff: Done

n1k0:
@celiacheff so I suggested a new module layout in 3c7f4eca1a1b55504c1e11f90db1be5a21960b91; basically the folder structure is now:

```
$ tree erp/imports
erp/imports
├── __init__.py
├── fetcher.py
├── importer.py
├── mapper
│   ├── __init__.py
│   ├── base.py
│   ├── gendarmerie.py
│   └── vaccination.py
└── printer.py
```

What do you think?

erp/mapper/gendarmerie.py
n1k0: Fixed in https://github.com/MTES-MCT/acceslibre/pull/182/commits/6b8668c0c1ed67154503a09786a8425186118e41

PR #196, opened 2021-06-11
User can change their email address

celiacheff:
[Trello](https://trello.com/c/keamDhZe/658-permettre-aux-utilisateurs-de-changer-ladresse-email-attach%C3%A9e-au-compte)


celiacheff:
Two security questions:
Should we ask to password before changing email ? (Solve forgotten opened sessions probelm)
Should the validate change endpoint be public or protected ? (Public: User can activate even if not logged in, maybe useful)

auth/views.py
```diff
@@ -0,0 +1,155 @@
...
+        form = forms.EmailChangeForm(initial={"email": request.user.email})
+    return render(
+        request,
+        "compte/mon_email.html",
+        context={"form": form},
+    )
+
+
+@login_required
+def change_email(request, activation_key):
```
n1k0, line 116: Quoting my global review comment:

> The issue with the validate endpoint being protected is that users must authenticate if they're not signed in already, and maybe they'll use their new email as username… which has not been updated nor activated yet in the db (that's why they reached this specific endpoint). So it MUST be public I think.

We should probably add a note in the activation email about that btw.

auth/views.py
```diff
@@ -0,0 +1,155 @@
...
+    if not activation_key:
+        return redirect("mon_email")
```
n1k0, line 118: We should probably alert users that something went wrong here.

core/urls.py
```diff
@@ -49,8 +48,10 @@ def test_sentry(request):
         CustomRegistrationView.as_view(form_class=CustomRegistrationForm),
         name="django_registration_register",
     ),
+    # TODO more things to move to auth
```
n1k0, line 51: Ideally we should list what needs to be moved in a Tech-Debt card if we land the patch as-is

tests/auth/test_auth.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # test email notification sent
+    assert len(mail.outbox) == 1
+    assert "Activation de votre compte" in mail.outbox[0].subject
+    assert new_email in mail.outbox[0].recipients()
+    assert "Vous avez demandé à changer votre email" in mail.outbox[0].body
+
+
+def test_user_validate_email_change(db, data):
+    uuid = "a603ae0a-4188-4098-99ca-3b853642c1c7"
```
n1k0, line 49: Here and more generally, I'd love we use a single, consistent term for activation tokens; how about `activation_token`?

n1k0 (commented):
@celiacheff nice work, still a few things to address and polish. To answer your [two questions](https://github.com/MTES-MCT/acceslibre/pull/196#issuecomment-859727300):

> Should we ask to password before changing email ? (Solve forgotten opened sessions probelm) 

This may be a nice to have later at some point (feel free to create a card), but let's start with the basics and land this thing first.

> Should the validate change endpoint be public or protected ? (Public: User can activate even if not logged in, maybe useful)

We can leave it public (unprotected) but we shouldn't sign users in automatically. So basically just after activating the new email, users should be offered a login form to be redirected to their account page, so… it's basically the same: they'll need to authenticate anyway. The only thing is that

The issue with the validate endpoint being protected is that users must authenticate if they're not signed in already, and maybe they'll use their new email as username… which has not been updated nor activated yet in the db (that's why they reached this specific endpoint). So it MUST be public I think.

auth/views.py
```diff
@@ -0,0 +1,150 @@
...
+
+def change_email(request, activation_token):
+    if not activation_token:
+        return redirect("mon_email")
+
+    user, failure = validate_from_token(activation_token)
+    if failure:
+        render(
+            request,
+            "compte/change_activation_failed.html",
```
n1k0, line 122: Nit: The template filename should be more explicit, eg. `email_change_activation_failed.html`

auth/views.py
```diff
@@ -0,0 +1,150 @@
...
+        change_message=f"Email modifié {user.email}",
+    )
+
+    messages.add_message(
+        request,
+        messages.SUCCESS,
+        "Votre email à été mis à jour avec succès !",
+    )
+
+    # return redirect("mon_compte")
```
n1k0, line 141: This comment can be safely removed

templates/compte/change_activation_success.html
```diff
@@ -0,0 +1,15 @@
...
+{% load static %}
+
+{% block page_title %}Activation réussie{% endblock%}
+{% block robots %}noindex, nofollow{% endblock %}
+
+{% block editorial_content %}
+<main role="main" id="content">
+    <div class="alert alert-success">
+      Votre adresse mail à été changée avec succès !
+      Vous pouvez vous connecter avec votre nouvelle adresse.
```
n1k0, line 12: Maybe link `vous connecter` to `{% url "login" %}`

templates/compte/mon_email.html
```diff
@@ -0,0 +1,24 @@
...
+    Changer mon email
+  </h3>
+
+  <form action="{% url "mon_email" %}" method="post">
+    <p>Votre email actuel est <strong>{{ request.user.email }}</strong>.</p>
+    {% csrf_token %}
+    {{ form|crispy }}
+
+    <p><button type="submit" class="btn btn-primary w-100">
+      Changer mon email
```
n1k0, line 21: Nit: *Mettre à jour mon adresse email*

n1k0 (commented):
Just a few more nits and remarks, and we're good to go

auth/management/commands/purge_tokens.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Command(BaseCommand):
+    help = "Supprime les demandes de changements d'emails non utilisés de plus de 24h"
+
+    def handle(self, *args, **options):
+        try:
+            EmailToken.objects.filter(expire_at__lt=datetime.now(timezone.utc)).delete()
+        except RuntimeError as err:
+            fatal(err)
```
n1k0, line 21: You can remove the try/catch here, as there's no chances a `RuntimeError` is raised for that call.

Also, makes me think that we should add a test [calling the command](https://www.kite.com/python/examples/5566/django-call-a-management-command-programmatically) over an expired entry in the database and assert that it's been properly deleted.

auth/forms.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def clean_email(self):
+        email1 = self.cleaned_data["email1"]
+        email2 = self.cleaned_data["email2"]
+
+        if email1 != email2:
+            raise ValidationError("Les emails ne correspondent pas")
+
+        if get_user_model().objects.filter(email__iexact=email1).count() > 0:
+            raise ValidationError("Veuillez choisir un email différent")
```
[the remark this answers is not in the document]
n1k0: So to check that the user has actually entered a different email that the one he currently uses, by adding a `user` dependency to `Form` constructor:

```py
class EmailChangeForm(forms.Form):
    email1 = define_email_field("Nouvel email")
    email2 = define_email_field("Confirmation nouvel email")
    
    def __init__(self, *args, user, **kwargs):
        self.user = user

    def clean(self):
        super().clean()
        email1 = self.cleaned_data.get("email1")
        email2 = self.cleaned_data.get("email2")

        if email1 == user.email:
            raise ValidationError("Vous devez fournir une nouvele adresse email")
        elif email1 != email2:
            raise ValidationError("Les emails ne correspondent pas")
        elif get_user_model().objects.filter(email__iexact=email1).count() > 0:
            raise ValidationError("Veuillez choisir un email différent")
        
        return email1
```

(sample code, untested)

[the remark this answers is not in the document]
celiacheff: That what it is doing. If the address is already in base, that's mean it's or the same than before, or already registered. i will make the message more explicit

auth/forms.py
```diff
@@ -0,0 +1,94 @@
...
+        if get_user_model().objects.filter(username__iexact=username).count() > 0:
+            raise ValidationError("Ce nom d'utilisateur est déjà pris.")
+        return username
+
+
+class EmailChangeForm(forms.Form):
+    email1 = define_email_field("Nouvel email")
+    email2 = define_email_field("Confirmation nouvel email")
+
+    def clean_email(self):
```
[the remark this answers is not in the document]
celiacheff: Fixed, forgot to edit after a change

auth/views.py
```diff
@@ -0,0 +1,155 @@
...
+    return render(request, "compte/index.html")
+
+
+@login_required
+def mon_identifiant(request):
+    if request.method == "POST":
+        form = forms.UsernameChangeForm(request.POST)
+        if form.is_valid():
+            username = form.cleaned_data["username"]
+            user = get_user_model().objects.get(id=request.user.id)
```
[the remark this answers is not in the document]
celiacheff: ^^ Fixed :)

n1k0 (approved):
👍

PR #201, opened 2021-06-15
Auto export datagouv

celiacheff:


erp/management/commands/export_to_datagouv.py
```diff
@@ -1,12 +1,50 @@
...
+
+
+def ping_mattermost(error=None):
+    if not settings.MATTERMOST_HOOK:
+        return
+    status = error if error else "Aucune erreur rencontrée"
+    requests.post(
+        settings.MATTERMOST_HOOK,
+        json={
+            "text": f"Export vers datagouv: {status}",
```
n1k0, line 48: Maybe add a link to the dataset for convenience

erp/management/commands/export_to_datagouv.py
```diff
@@ -1,12 +1,50 @@
...
+        )
+
     def handle(self, *args, **options):
-        try:
-            export_to_datagouv.job()
-            print("Data exported successfully to 'export.csv'")
-        except KeyboardInterrupt:
-            print("Interrompu.")
+        verbose = options.get("verbose", False)
+        csv_path = "acceslibre.csv"
```
n1k0, line 22: How about creating the export file in a tempfile? https://docs.python.org/3.8/library/tempfile.html

erp/management/commands/export_to_datagouv.py
```diff
@@ -1,12 +1,50 @@
...
+        with open(csv_path, "w", newline="") as csv_file:
+            try:
+                erps = Erp.objects.published().having_a11y_data()
+                export_schema_to_csv(csv_file, erps, EtalabMapper)
+                if verbose:
+                    print("Local export successful: 'acceslibre.csv'")
+                if settings.DATAGOUV_API_KEY:
```
n1k0, line 29: Here's a general idea: how about we only publish to datagouv if the `DATAGOUV_API_KEY` is defined? that way we don't need to declare environments, just setting an env var or not.

erp/management/commands/export_to_datagouv.py
celiacheff: The problem is that when you upload a file, the ressource title is changed to the file name (same with the website). And I can't name a tmp with the tempfile module (yeah NamedTemporaryFile doesn't mean you can choose a name)

n1k0: Maybe you can create a tmp directory and create the file in it? 🤔

core/settings.py
```diff
@@ -32,11 +32,13 @@ def get_env_variable(var_name, required=True, type=str):
         return os.environ.get(var_name)
 
 
+ENVIRONMENT = "dev"
 STAGING = False
 SITE_NAME = "acceslibre"
 SITE_HOST = "acceslibre.beta.gouv.fr"
 SITE_ROOT_URL = f"https://{SITE_HOST}"
 SECRET_KEY = get_env_variable("SECRET_KEY")
+DATAGOUV_API_KEY = get_env_variable("DATAGOUV_API_KEY")
```
n1k0, line 41: Don't forget to update `README.md` and `.env.sample` files accordingly.

core/settings.py
```diff
@@ -38,6 +38,8 @@ def get_env_variable(var_name, required=True, type=str):
 SITE_ROOT_URL = f"https://{SITE_HOST}"
 SECRET_KEY = get_env_variable("SECRET_KEY")
 DATAGOUV_API_KEY = get_env_variable("DATAGOUV_API_KEY", required=False)
+DATAGOUV_DOMAIN = "https://demo.data.gouv.fr"
```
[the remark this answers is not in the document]
celiacheff: I need it to display the link to the dataset in the Mattermost report. But I could do otherwise, like providing only the production link, so I could put only the api url in settings. Tell me what you think about this.

erp/export/export.py
```diff
@@ -35,8 +34,22 @@ def upload_to_datagouv(csv_path):
...
+        response = requests.post(
+            url,
+            files={"file": open(csv_path, "rb")},
+            headers={"X-API-KEY": settings.DATAGOUV_API_KEY},
+        )
+        response.raise_for_status()
+    except (
+        requests.exceptions.RequestException,
+        requests.exceptions.Timeout,
+        json.JSONDecodeError,
```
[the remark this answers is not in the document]
celiacheff: The call we just below, I moved it.

core/settings.py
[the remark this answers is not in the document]
n1k0: Ah yes I missed it, keep it that way it's fine

erp/export/export.py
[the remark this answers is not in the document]
n1k0: Then we now need to catch `AssertionError` ;)

erp/management/commands/start_scheduler.py
```diff
@@ -12,9 +13,12 @@ def add_arguments(self, parser):
     def handle(self, *args, **options):
         schedule.every().day.at("04:00").do(call_command, "purge_inactive_accounts")
         schedule.every().hour.do(call_command, "purge_tokens")
+        if settings.DATAGOUV_API_KEY:
+            schedule.every().day.at("01:00").do(call_command, "export_to_datagouv")
         schedule.every(3).hours.do(call_command, "notify_changed_erps", hours=3)
         schedule.every().hour.do(call_command, "import_dataset", "vaccination")
         schedule.every().day.do(call_command, "import_dataset", "gendarmerie")
+        schedule.every().day.at("01:00").do(call_command, "export_to_datagouv")
```
n1k0, line 21: Why do we schedule `export_to_datagouv` twice? 

Also I think we shouldn't need testing for the `DATAGOUV_API_KEY` presence in the scheduler, this check should rather be done in the command itself.

erp/management/commands/export_to_datagouv.py
```diff
@@ -1,12 +1,61 @@
...
+            if res:
+                ping_mattermost(len(erps))
+                self.log("Dataset uploaded")
+        except RuntimeError as err:
+            ping_mattermost(error=str(err))
+            raise err
+
+
+def ping_mattermost(count=0, error=None):
+    if not settings.MATTERMOST_HOOK and not settings.DATAGOUV_API_KEY:
```
n1k0, line 47: I think you want an `or` statement here.

.circleci/config.yml
```diff
@@ -14,6 +14,7 @@ jobs:
           EMAIL_HOST_USER: xxx
           EMAIL_HOST_PASSWORD: xxx
           SCALINGO_APP: access4all
+          DATAGOUV_API_KEY: xxx
```
n1k0, line 17: Should we set this in the test needing it instead? It's a bit surprising an optional env var is required in CI…

tests/erp/export/test_export_unit.py
```diff
@@ -39,23 +42,22 @@ def example_data(db) -> List[Erp]:
...
-    assert erp_1["transport_station_presence"] == str(
-        mapped_data[1].transport_station_presence
-    )
+    with open(file, "r") as csv_file:
+        reader = csv.DictReader(csv_file, fieldnames=first_row)
+        next(reader)  # Skip headers
 
-    # Validate data with schema
+        erp_0 = next(reader)
+        print(erp_0)
```
n1k0, line 53: We may want to remove this debug leftover

erp/export/export.py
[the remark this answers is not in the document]
n1k0: @celiacheff would you mind adding an `except` clause for assertion errors? so we can land the patch. thanks.

n1k0 (approved):
👍

PR #203, opened 2021-06-15
Remove all siret-based activity status checks.

n1k0:
[Trello card](https://trello.com/c/zmUoGJR1/664-suppression-fonctionnalit%C3%A9s-li%C3%A9es-au-siret)


celiacheff (approved):
Less is more :)

PR #206, opened 2021-06-16
Deduplicate preexisting and imported gendarmeries.

n1k0:
@celiacheff We have preexisting "gendarmerie" duplicates, and we could check and remove them at scheduled import time. This patch is a basic PoC.


n1k0:
Sample output:

image

n1k0:
Now the question is, how could we make this CLEAN? 😓

celiacheff (approved):

PR #209, opened 2021-06-21
Change etalab's schema to lon/lag instead of geopoint

celiacheff:
[Trello](https://trello.com/c/E3KdpOgT/678-modifer-les-coordonn%C3%A9es-du-schema-en-champs-lon-lat-s%C3%A9par%C3%A9s)


erp/management/commands/export_to_datagouv.py
```diff
@@ -22,18 +22,28 @@ def add_arguments(self, parser):
...
+        local = options.get("local", False)
+
         csv_path = "acceslibre.csv"
         self.log("Starting export")
         try:
             erps = Erp.objects.published().having_a11y_data()
             export_schema_to_csv(csv_path, erps, EtalabMapper)
             self.log("Local export successful: 'acceslibre.csv'")
+            if local:
+                return
```
n1k0, line 46: How about a more classic `--dry-run` option?

n1k0 (approved):
👍  with a minor remark addressed and fixed tests

erp/management/commands/export_to_datagouv.py
celiacheff: I want to run it and create the export, but not to upload it. It helps me to update the samples data required for testing. Maybe a `--no-upload` or `--local-only` flag maybe more expressive ?

n1k0: `--skip--upload` sounds good

tests/erp/export/test_export_integration.py
```diff
@@ -21,13 +20,6 @@ def test_csv_creation(db):
         export_schema_to_csv(dest_path, erps, EtalabMapper)
 
         assert Path(dest_path).exists() is True
-
-        resource = Resource(
-            "erp/export/static/exemple-valide.csv",
-            schema="erp/export/static/schema.json",
-        )
-        result = validate_resource(resource)
-        assert result.get("errors") == []
```
n1k0, line 30: Why is this test removed?

celiacheff: I made a separated test for this at the bottom of the file. It wasn't the scope of that test at all, since the generated file wasn't tested against the ressource anyway.

n1k0: Oops indeed, sorry for the nois

celiacheff:
> In France longitude are usually between `2` and `4`, we should rather feature an example reflecting that fact, eg. `2.352222` for Paris, with a latitude of `48.856613` — these two values would be great examples methinks.

Good idea 

> Also, why aren't these actual float?

You mean in the mapper ? Yeah, that doesn't influence the result, but it'll be more clear

n1k0 (approved):

PR #213, opened 2021-06-23
Allow users to delete their account.

celiacheff:
[Trello](https://trello.com/c/Fnq4H0P0/416-fonctionnalit%C3%A9-suppression-de-compte)


compte/views.py
```diff
@@ -149,6 +154,30 @@ def change_email(request, activation_token):
...
+        )
+        return redirect("mon_compte")
+
+    if res is True:
+        logout(request)
+
+        messages.add_message(
+            request,
+            messages.SUCCESS,
+            "Votre compte à bien été désactivé",
```
n1k0, line 175: Supprimé plutôt que désactivé

Pipfile
```diff
@@ -49,6 +49,7 @@ schedule = "*"
 sentry-sdk = "*"
 uritemplate = "*"
 uuid = "*"
+shortuuid = "*"
```
n1k0, line 52: Do we really need a package to generate a random string?

compte/management/commands/purge_inactive_accounts.py
```diff
@@ -39,7 +39,7 @@ def handle(self, *args, **options):
             get_user_model()
             .objects.annotate(erps_count=Count("erp"))
             .filter(
-                is_active=False,
+                last_login=None,
```
n1k0, line 42: Have you considered using a soft-delete package like this one? https://pypi.org/project/django-soft-delete/

n1k0:
After having deleted my account, erps I added shows this:

image

I'd rather see `<compte supprimé>` or similar.

compte/views.py
```diff
@@ -149,6 +154,30 @@ def change_email(request, activation_token):
...
+    except RuntimeError as err:
+        messages.add_message(
+            request,
+            messages.WARNING,
+            f"Erreur lors de la désactivation du compte: {err}",
+        )
+        return redirect("mon_compte")
+
+    if res is True:
+        logout(request)
```
n1k0, line 170: The return value of `remove_personal_informations` is always `True`, is this check necessary?

compte/views.py
```diff
@@ -149,6 +154,30 @@ def change_email(request, activation_token):
...
 
+@login_required
+def delete_account(request):
+    try:
+        res = remove_personal_informations(request.user, shortuuid.uuid)
+    except RuntimeError as err:
+        messages.add_message(
+            request,
+            messages.WARNING,
+            f"Erreur lors de la désactivation du compte: {err}",
```
n1k0, line 165: How about not rendering the technical exception message to the end user? And how about logging the error instead, so Sentry can be notified?

subscription/management/commands/notify_changed_erps.py
```diff
@@ -54,7 +54,11 @@ def get_notifications(self, hours, now=None):
             erp = versioning.extract_online_erp(version)
             if not erp:
                 continue
-            subscribers = [sub.user for sub in ErpSubscription.objects.subscribers(erp)]
+            subscribers = [
+                sub.user
+                for sub in ErpSubscription.objects.subscribers(erp)
+                if sub.user.is_active
```
n1k0, line 60: We should rather enforce the `is_active` filter in the `ErpSubscription.objects.subscribers` queryset method.

n1k0:
Now wondering out loud: why don't we just delete the user record from the database, benefiting from all the defined integrity constraints? What are the risks we're preventing keeping these anonymized records? Why are we even bothering defining `ON DELETE CASCADE` or `SET NULL` constraints if we never leverage them? 🤔

n1k0:
@celiacheff Hmm there's a big bug:

- Log in
- Go to "Mon compte"
- Click on "Supprimer mon compte [Supprimer]"
- Click on "Retour" (cancel the operation)
- **The account is deleted**

templates/compte/delete_account_warning.html
```diff
@@ -0,0 +1,28 @@
...
+    <div class="d-flex justify-content-between">
+      <span>
+        <a class="btn btn-sm btn btn-outline-info text-nowrap" href="{% url "delete_account" %}">
+         Retour
+        </a>
+      </span>
+      <span>
+        <a class="btn btn-sm btn-danger text-nowrap" href="{% url "delete_account" %}">
+         Confirmer la suppression
+        </a>
```
n1k0, line 24: This should definitely NOT be a link, rather a button which submits a form in POST, with csrf checks and so on. Also btw, the two "Retour" and "Confirmer la suppression" links both point to the same deletion URL, so that's why when clicking "Retour" the account is actually deleted.

compte/views.py
```diff
@@ -149,6 +154,30 @@ def change_email(request, activation_token):
         )
 
 
+@login_required
+def delete_account(request):
+    try:
+        res = remove_personal_informations(request.user, shortuuid.uuid)
+    except RuntimeError as err:
```
n1k0, line 161: We must ensure that the request has been submitted in `POST`, ideally validating a form containing a boolean field checked expressing user consent before actually soft-deleting the user account.

n1k0:
Replying to myself:

> Now wondering out loud: why don't we just delete the user record from the database

Unfortunately, django-reversion serialized records still references old user primary keys, so hard-deleting a User is not an option. Soft-deletion it is then.

compte/management/commands/purge_inactive_accounts.py
celiacheff: That raise an interesting point: should we consider the user as deleted, or as an anonymized user. The is_active field was used here as an non-actived account, but to it's not the primary use

n1k0: What is the primary usage of the `is_active` field? 🤔

templates/compte/delete_account_warning.html
n1k0: @celiacheff if that helps, it's basically exactly what has been done for erp deletion https://github.com/MTES-MCT/acceslibre/blob/master/erp/forms.py#L481-L493

Pipfile
n1k0: For the records, generating a 10c random string can be achieved like this using stlib:

```py
import random
import string

"".join(random.choice(string.ascii_letters) for i in range(10))
```

tests/compte/test_remove_account.py
```diff
@@ -0,0 +1,65 @@
...
+    client.force_login(data.niko)
+
+    response = client.post(
+        reverse("delete_account"),
+        data={"confirm": False},
+        follow=True,
+    )
+
+    assert response.status_code == 200
+    assert response.wsgi_request.user == data.niko
```
n1k0, line 52: How about asserting the context form contains the expected error, eg using 

```py
assert "confirm" in response.context["form"].errors
```

compte/service.py
```diff
@@ -59,3 +60,19 @@ def validate_from_token(activation_token, today=datetime.now(timezone.utc)):
     email_token.delete()
 
     return user, None
+
+
+def remove_personal_informations(user, uuid_gen):
+    user.email = ""
+    user.username = "<user-deleted>"
```
n1k0, line 67: If we absolutely want to stick with English here, we should update the `format_username` templatetag to handle this value and expose a French version of it.

Pipfile
n1k0: Haha, we even already have a `random_string` function in `core.lib.text` :)

PR #216, opened 2021-06-25
Parallelize python tests.

n1k0:
From https://schegel.net/posts/making-your-django-tests-faster/


n1k0:
Doesn't work on circleci, closing.

PR #219, opened 2021-06-30
Notify users when they have pending unpublished erps

celiacheff:
[Trello](https://trello.com/c/TbIsTeUx/276-en-tant-quadministrateur-je-veux-pouvoir-relancer-les-contributeurs-nayant-pas-encore-publi%C3%A9-leur-fiche)


compte/migrations/0003_auto_20210629_1730.py
```diff
@@ -0,0 +1,21 @@
+from django.contrib.auth import get_user_model
+from django.db import migrations
+
+from compte.models import UserPreferences
+
+
+def migrate_prestations_to_commentaire(apps, schema_editor):
```
n1k0, line 7: This function name doesn't match what it does :)

erp/management/commands/notify_unpublished_erps.py
```diff
@@ -0,0 +1,84 @@
...
+    def handle(self, *args, **options):
+        notifications = self.get_notifications()
+        total = len(notifications)
+        sent_ok = 0
+        for notification in notifications:
+            sent_ok += 1 if self.send_notification(notification) else 0
+        if total > 0:
+            plural = "s" if total > 1 else ""
+            mattermost.send(
+                f"{sent_ok}/{total} relance{plural} d'ERPs non-publiés",
```
n1k0, line 45: We should use `plural` for `non-publiés` as well.

erp/management/commands/notify_unpublished_erps.py
```diff
@@ -0,0 +1,84 @@
...
+            published=False,
+            updated_at__lte=self.now - timedelta(days=7),
+            user__isnull=False,
+        )
+
+        for erp in erps:
+            try:
+                if self._should_notify(erp.user) and self._due_time(erp):
+                    notifications.append((erp.user, erp))
+            except User.DoesNotExist:
```
n1k0, line 75: How could this happen?

erp/management/commands/notify_unpublished_erps.py
```diff
@@ -0,0 +1,84 @@
...
+                continue
+
+        return notifications
+
+    def _should_notify(self, user):
+        return UserPreferences.objects.get(user=user).notify_on_unpublished_erps is True
```
n1k0, line 81: This may throw if no record is associated to this user yet. Oh. That's why you're catching `DoesNotExist` above… I'd rather we return `False` if no object is found. Also, this screams to have a `user.preferences` getter, which we probably have as it's a model fk…

erp/management/commands/start_scheduler.py
```diff
@@ -23,6 +23,7 @@ def handle(self, *args, **options):
         schedule.every(3).hours.do(call_command, "notify_changed_erps", hours=3)
         schedule.every().hour.do(call_command, "import_dataset", "vaccination")
         schedule.every().day.do(call_command, "import_dataset", "gendarmerie")
+        schedule.every().day.at("9:30").do(call_command, "notify_unpublished_erps")
```
n1k0, line 26: Should we send a reminder every day? We're going to be flagged as spam rather quickly… How about once a week? or even once a month?

templates/compte/notifications_disabled.html
```diff
@@ -0,0 +1,16 @@
...
+{% load static %}
+
+{% block page_title %}Désactivations des rappels{% endblock %}
+{% block robots %}noindex, nofollow{% endblock %}
+
+{% block editorial_content %}
+  <main role="main" id="content">
+    <h3 class="mb-3">Désabonnement</h3>
+    <div class="alert alert-info">
+      Votre demande a bien été prise en compte.
```
n1k0, line 13: I'd rather say *Vos préférences ont bien été mises à jour*, so we can have a generic, reusable `preferences_updated.html` template for future other preferences.

Also, why didn't you use flash messages?

compte/views.py
```diff
@@ -260,3 +261,15 @@ def mes_abonnements(request):
...
+@login_required
+def disable_reminders(request):
+    prefs = UserPreferences.objects.get(user=request.user)
+    prefs.notify_on_unpublished_erps = False
+    prefs.save()
+
+    return render(
+        request,
+        "compte/notifications_disabled.html",
+    )
```
n1k0, line 275: I know this is more work, and we only have a single preference right now, but how about having a preferences management page with a dedicated form instead? 🤔

erp/management/commands/notify_unpublished_erps.py
celiacheff: Too much refactoring -_-

erp/management/commands/start_scheduler.py
celiacheff: That's why I have a _due_time() guard. The user is notified only once per week (day diff mod 7) after the last updated_at, if the ERP is not published.

n1k0: Ah yes indeed I reviewed to fast :D

templates/compte/notifications_disabled.html
celiacheff: Well, the task was asking this, but thats makes more sens especially with your next remark

compte/views.py
celiacheff: Yeah, I started with that, but then I checked what others do with the simple case, and I took the simplest path. But I'll do that

erp/management/commands/start_scheduler.py
celiacheff: Every 7 days to be precise

erp/management/commands/notify_unpublished_erps.py
celiacheff: The DoesNotExist catch if for the case no user is registered for this erp, they MUST have a preference anyway

n1k0: I was thinking `UserPreferences.objects.filter(user=user).first()`.

erp/management/commands/notify_unpublished_erps.py
n1k0: Reopening this comment because the situation where a user doesn't exist can't really happen, as you filter erps specifying `user__inull=False`, so there's ALWAYS an existing `erp.user` passed to `_should_notify`.

n1k0 (commented):
Better than adding comments here and there, here's a patch of what I think should be addressed in the patch before landing:

```diff
diff --git a/compte/models.py b/compte/models.py
index 81a1bb95..794b4084 100644
--- a/compte/models.py
+++ b/compte/models.py
@@ -1,5 +1,5 @@
 from django.conf import settings
-from django.contrib.auth.models import User
+from django.contrib.auth import get_user_model
 from django.db import models
 from django.db.models.signals import post_save
 from django.dispatch import receiver
@@ -41,7 +41,7 @@ class UserPreferences(models.Model):
         verbose_name="Recevoir des mails de rappel de publication",
     )
 
-    @receiver(post_save, sender=User)
+    @receiver(post_save, sender=get_user_model())
     def save_profile(sender, instance, created, **kwargs):
         if created:
             user_prefs = UserPreferences(user=instance)
diff --git a/compte/urls.py b/compte/urls.py
index b0ae72e9..94954828 100644
--- a/compte/urls.py
+++ b/compte/urls.py
@@ -26,12 +26,12 @@ urlpatterns = [
         name="mes_contributions_recues",
     ),
     path(
-        "suppression",
+        "suppression/",
         views.delete_account,
         name="delete_account",
     ),
     path(
-        "preferences",
+        "preferences/",
         views.mes_preferences,
         name="mes_preferences",
     ),
diff --git a/erp/management/commands/notify_unpublished_erps.py b/erp/management/commands/notify_unpublished_erps.py
index 26296541..b2e22cef 100644
--- a/erp/management/commands/notify_unpublished_erps.py
+++ b/erp/management/commands/notify_unpublished_erps.py
@@ -1,9 +1,9 @@
 import logging
+
 from datetime import datetime, timezone, timedelta
 from typing import List
 
 from django.conf import settings
-from django.contrib.auth.models import User
 from django.core.management.base import BaseCommand
 
 from core import mailer, mattermost
@@ -68,11 +68,8 @@ class Command(BaseCommand):
         )
 
         for erp in erps:
-            try:
-                if self._should_notify(erp.user) and self._due_time(erp):
-                    notifications.append(erp)
-            except User.DoesNotExist:
-                continue
+            if self._should_notify(erp.user) and self._due_time(erp):
+                notifications.append(erp)
 
         return notifications
 
diff --git a/erp/templatetags/a4a.py b/erp/templatetags/a4a.py
index 1ddc5112..8664f535 100644
--- a/erp/templatetags/a4a.py
+++ b/erp/templatetags/a4a.py
@@ -32,6 +32,7 @@ def active_compte_section(path, test):
             test == "mon_email" and path.startswith("/compte/email/"),
             test == "mes_contributions" and path.startswith("/compte/contributions/"),
             test == "mes_abonnements" and path.startswith("/compte/abonnements/"),
+            test == "mes_preferences" and path.startswith("/compte/preferences/"),
             test == "mot_de_passe"
             and path
             in [
diff --git a/templates/compte/base.html b/templates/compte/base.html
index 3a0273e6..5b4b05ac 100644
--- a/templates/compte/base.html
+++ b/templates/compte/base.html
@@ -43,6 +43,13 @@
           Mes abonnements
         </a>
       </li>
+      <li class="nav-item">
+        <a class="nav-link text-nowrap {{ request.path|active_compte_section:"mes_preferences" }}"
+          href="{% url "mes_preferences" %}">
+          <i aria-hidden="true" class="icon icon-pencil mr-1"></i>
+          Mes préférences
+        </a>
+      </li>
       <li class="nav-item">
         <a class="nav-link text-nowrap {{ request.path|active_compte_section:"mon_identifiant" }}"
         href="{% url "mon_identifiant" %}">
diff --git a/templates/compte/mes_preferences.html b/templates/compte/mes_preferences.html
index 94366e35..08379f91 100644
--- a/templates/compte/mes_preferences.html
+++ b/templates/compte/mes_preferences.html
@@ -1,12 +1,12 @@
-{% extends "editorial/base.html" %}
-{% load crispy_forms_filters %}
+{% extends "compte/base.html" %}
 
+{% load crispy_forms_filters %}
 {% load static %}
 
-{% block page_title %}Compte - Préférences{% endblock %}
+{% block page_title %}Préférences{% endblock %}
 {% block robots %}noindex, nofollow{% endblock %}
 
-{% block editorial_content %}
+{% block compte_content %}
   <main role="main" id="content">
     <h3 class="mb-3">Mes préférences</h3>
     <form action="{% url "mes_preferences" %}" method="post">
```

erp/management/commands/notify_unpublished_erps.py
```diff
@@ -0,0 +1,83 @@
...
+            published=False,
+            updated_at__lte=self.now - timedelta(days=7),
+            user__isnull=False,
+        )
+
+        for erp in erps:
+            try:
+                if self._should_notify(erp.user) and self._due_time(erp):
+                    notifications.append(erp)
+            except User.DoesNotExist:
```
n1k0, line 74: Reopening this comment because the situation where a user doesn't exist can't really happen, as you filter erps specifying `user__inull=False`, so there's ALWAYS an existing `erp.user` passed to `_should_notify`.

celiacheff: I know but I have cases where the ERP have an user, but erp.user still throw. You should have some of them in your db, remove the catch and run `./manage.py notify_unpublished_erps`. Haven't found why yet.

n1k0 (approved):

PR #229, opened 2021-07-09
Change wording email notifications

z3ntUx:


core/lib/diff.py
```diff
@@ -5,11 +5,6 @@ def dict_diff_keys(old, new):
     diff = []
     for key, oldval in old.items():
         newval = new.get(key)
-        # Convert points to tuples of float to avoid GEOS ParseException
-        if isinstance(oldval, Point):
-            oldval = oldval.coords
-        if isinstance(newval, Point):
-            newval = newval.coords
```
n1k0, line 12: Build fails with a flake8 error: `./core/lib/diff.py:1:1: F401 'django.contrib.gis.geos.Point' imported but unused`. Indeed, we removed code referencing `Point`, which is not used anymore in this module.

Hint: add flake8 support to your Pycharm installation so you can see these errors before even pushing to git.

n1k0 (approved):
That was fun ☺️

n1k0 (approved):

PR #283, opened 2022-11-16
Fix accessibilite, add some tests.

mlvernay:


erp/views.py
```diff
@@ -871,13 +871,16 @@ def contrib_a_propos(request, erp_slug):
...
+            accessibilite = form.save(commit=False)
+            accessibilite.erp = erp
+            accessibilite.save()
             if request.user.is_authenticated and erp.user is None:
                 erp.user = request.user
                 erp.save()
                 if form.cleaned_data["subscribe"] is True:
                     ErpSubscription.subscribe(erp, request.user)
+            else:
+                erp.save()
```
mlvernay, line 883: This `else`, is here to avoid a double save in db. 
Before my change we were doing a first `erp.save()` and a second one to save the user 

My guess is that we could also do as following, as `ErpSubscription.subscribe` does not require a erp.user, WDYT ?
```
            if request.user.is_authenticated and erp.user is None:
                erp.user = request.user
                if form.cleaned_data["subscribe"] is True:
                    ErpSubscription.subscribe(erp, request.user)
             erp.save()
```

z3ntUx (approved):

PR #286, opened 2022-11-22
Fix global search when we have large results.

mlvernay:
:warning: double check to be ISO functional with previous version

On a global search made on "restaurant" "Paris", it tooked ~45s, now taking  ~3 s

### DIFF
--- a/erp/provider/acceslibre.py
+++ b/erp/provider/acceslibre.py
@@ -1,43 +1,42 @@
 import logging
 
+from django.forms.models import model_to_dict
+
 from erp.models import Erp
 
 logger = logging.getLogger(__name__)
 
 
-def _check_doublons(erp, other):
-    return all(
-        [erp.get("siret") != other.get("siret"), erp.get("nom") != other.get("nom")]
-    )
+def _not_duplicate(erp, other):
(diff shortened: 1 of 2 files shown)


erp/provider/acceslibre.py
```diff
@@ -1,43 +1,42 @@
...
 
 logger = logging.getLogger(__name__)
 
 
-def _check_doublons(erp, other):
-    return all(
-        [erp.get("siret") != other.get("siret"), erp.get("nom") != other.get("nom")]
-    )
+def _not_duplicate(erp, other):
+    return all([erp.get("siret") != other.get("siret"), erp.get("nom") != other.get("nom")])
```
mlvernay, line 11: Just renamed the func + lint

erp/provider/acceslibre.py
```diff
@@ -1,43 +1,42 @@
...
+                    "user": erp.user_id,
+                    "published": erp_as_dict["published"],
+                    "icon": erp_as_dict["icon"],
+                    "activite": erp_as_dict["activite"],
+                },
+                source="acceslibre",
+                coordonnees=erp_as_dict["geom"],
+                naf=None,
+                **erp_as_dict
+            )
```
mlvernay, line 40: We are now looping only once through the queryset results + not regetting the erp from db in each iteration

PR #334, opened 2022-12-22
Draft: Write API

mlvernay:


mlvernay:
Temporarely include "API-Key mandatory", do not merge before clients integration

PR #442, opened 2023-03-20
Feat/api key mandatory

mlvernay:


mlvernay:
Do not merge before clients integration

PR #540, opened 2023-05-17
Bring back Crisp.

mlvernay:
Cannot find any trace of a previous integration.


Anto59290 (approved):
Apart from the discussion we had on GDPR conformity to be confirmed it is good for me

PR #551, opened 2023-05-22
Improve search loc

Anto59290:
- [x] Revert some changes done by black
- [x] Revert template linting
- [x] Remove all Todo
- [x] Check tests / add tests
- [x] Check if we need to add back select related


erp/views.py
```diff
@@ -235,44 +248,69 @@ def _get_search_qs_and_context(commune_slug, lat, lon, what, where):
...
+                "pager": pager,
+            }
+        )
+
+    return render(request, "search/results.html", context=context)
 
 
 @waffle_switch("USE_GLOBAL_MAP")
 def global_map(request, commune_slug=None):
+    # TODO understand where is this map / what is the purpose
```
Anto59290, line 306: @mlvernay If we can discuss this tomorrow

erp/views.py
```diff
@@ -193,104 +193,121 @@ def _update_search_pager(pager, commune):
...
+    return {
+        k: "" if get_parameters.get(k, "") == "None" else get_parameters.get(k, "")
+        for k, v in get_parameters.items()
+        if k in allow_list
+    }
 
 
 def _parse_location_or_404(lat, lon):
-    if not lat or not lon:
+    if not all([lat, lon]):
```
mlvernay, line 206: approx 4 to 10 times slower than the previous `or`. It is not a big deal, but maybe better to keep it as it was no ?

erp/views.py
```diff
@@ -193,104 +193,127 @@ def _update_search_pager(pager, commune):
...
+def _filter_erp_by_location(queryset, **kwargs):
+    search_type = kwargs.get("search_type")
+
+    # TODO setting for this ?
+    if search_type == "municipality":
+        return queryset.filter(commune_ext=kwargs.get("municipality"))
+
+    if search_type == settings.ADRESSE_DATA_GOUV_SEARCH_TYPE_CITY:
+        postcode = kwargs.get("postcode")
+        return queryset.filter(commune_ext__code_postaux__contains=[postcode])
```
Anto59290, line 223: Do this in 2 steps, retrieve commune from post code and then use the commune name in the ERP

erp/views.py
```diff
@@ -193,104 +193,127 @@ def _update_search_pager(pager, commune):
...
-        "lon": lon,
-        "what": what,
-        "where": where,
-        "commune_json": commune.toTemplateJson() if commune else None,
+def _filter_erp_by_location(queryset, **kwargs):
+    search_type = kwargs.get("search_type")
+
+    # TODO setting for this ?
+    if search_type == "municipality":
+        return queryset.filter(commune_ext=kwargs.get("municipality"))
```
Anto59290, line 219: TODO check this use of the FK

erp/provider/geocoder.py
```diff
@@ -32,6 +32,7 @@ class Provider(StrEnum):
 ]
 
 
+# TODO check if this can be deleted
```
Anto59290, line 35: @mlvernay I am not sure how we could end up in this code / if that was even possible. We can discuss it next week. I removed the final usage of this function, only a few tests are left. Let me know if:
- You are ok with me removing this (I can show you my doubts about the usage)
- You would prefer this to be removed here or in a separate PR

erp/views.py
```diff
@@ -193,104 +193,129 @@ def _update_search_pager(pager, commune):
...
+    search_type = kwargs.get("search_type")
+
+    # TODO setting for this ?
+    if search_type == "in_municipality":
+        return queryset.filter(commune=kwargs.get("municipality").nom)
+
+    if search_type == settings.ADRESSE_DATA_GOUV_SEARCH_TYPE_CITY:
+        return queryset.filter(code_postal=kwargs.get("postcode"))
+
+    # TODO handle trad with ML + setting for this
```
Anto59290, line 226: @mlvernay I know you worked on translations recently, could we talk 5 minutes about how the translations are handled between backend and frontend ? I would love to remove those hardcoded string but I am not sure about the process

erp/views.py
```diff
@@ -206,91 +214,102 @@ def _parse_location_or_404(lat, lon):
...
+        return queryset.annotate(city=Lower("commune")).filter(city=kwargs.get("municipality").lower())
+
+    if search_type in (settings.ADRESSE_DATA_GOUV_SEARCH_TYPE_HOUSENUMBER, "Autour de moi", translate("Autour de moi")):
+        lat, lon = kwargs.get("lat"), kwargs.get("lon")
+        location = _parse_location_or_404(lat, lon)
+        return queryset.nearest(location, max_radius_km=0.2)
+
+    if search_type == settings.ADRESSE_DATA_GOUV_SEARCH_TYPE_STREET:
+        street_name = kwargs.get("street_name")
+        return queryset.filter(code_postal=kwargs.get("postcode"), voie__icontains=street_name)
```
mlvernay, line 233: why using a var for street_name and not for postcode? Not a pb at all.

erp/views.py
```diff
@@ -206,91 +214,102 @@ def _parse_location_or_404(lat, lon):
...
+    return queryset
+
+
+def search(request):
+    filters = _cleaned_search_params_as_dict(request.GET)
+    base_queryset = Erp.objects.published().search_what(filters.get("what"))
+    queryset = _filter_erp_by_location(base_queryset, **filters)
+
+    paginator = Paginator(queryset, 10)
+    pager = paginator.get_page(request.GET.get("page", 1))
```
mlvernay, line 243:
```suggestion
    pager = paginator.get_page(request.GET.get("page") or 1)
```
to avoid the case "page=0"

erp/views.py
```diff
@@ -206,91 +214,102 @@ def _parse_location_or_404(lat, lon):
...
+def search_in_municipality(request, commune_slug):
+    municipality = get_object_or_404(Commune, slug=commune_slug)
+    filters = _cleaned_search_params_as_dict(request.GET)
+    filters["search_type"] = settings.IN_MUNICIPALITY_SEARCH_TYPE
+    filters["municipality"] = municipality
+    base_queryset = Erp.objects.published().search_what(filters.get("what"))
+    queryset = _filter_erp_by_location(base_queryset, **filters)
+
+    paginator = Paginator(queryset, 10)
     pager = paginator.get_page(request.GET.get("page", 1))
```
mlvernay, line 265: same here

erp/views.py
```diff
@@ -206,91 +214,102 @@ def _parse_location_or_404(lat, lon):
...
 
 
-def search(request, commune_slug=None):
-    where, what, lat, lon, code = _clean_search_params(request.GET, "where", "what", "lat", "lon", "code")
-    where = where or "France entière"
-    qs, context = _get_search_qs_and_context(commune_slug, lat, lon, what, where)
-    # pager
-    paginator = Paginator(qs, 10)
+def search_in_municipality(request, commune_slug):
+    municipality = get_object_or_404(Commune, slug=commune_slug)
```
mlvernay, line 257: Not sure we should based our search on this model. This supposed to have an up-to-date Commune database, which is not the case atm.

mlvernay (commented):
First review, I will now make some functionnal tests.

mlvernay (commented):
First review, I will now make some functionnal tests.

erp/views.py
Anto59290: I will change it to make it more consistent

erp/views.py
Anto59290: I mostly rewrote what was already done I believe for this view. My proposition would be to analyze the issue further and create a ticket to check if we should:
- Try to remove this model
- Try to update the data frequently
- Leave everything as is

mlvernay (approved):

PR #558, opened 2023-05-24
Translate free texts.

mlvernay:
Basic server side translation for free texts.
This s2s solution avoid a malicious usage of our key/translation service. This is a naive approach which can be enhanced if needed, especially with some cache mechanisms.


erp/views.py
```diff
@@ -234,6 +235,19 @@ def _get_search_qs_and_context(commune_slug, lat, lon, what, where):
     }
 
 
+def _translate_erp(erp: Erp, target_lang: str):
```
Anto59290, line 238: I think could have been a method of the erp object no ? This could ease the use of translated fields elsewhere (exports, api, etc).

Anto59290 (approved):

PR #559, opened 2023-05-24
Use django-environ

mlvernay:


core/settings.py
```diff
@@ -255,15 +254,15 @@ def expose_site_context(request):
...
-EMAIL_PORT = get_env_variable("EMAIL_PORT", type=int)
-EMAIL_HOST_USER = get_env_variable("EMAIL_HOST_USER")
-EMAIL_HOST_PASSWORD = get_env_variable("EMAIL_HOST_PASSWORD")
+EMAIL_HOST = env("EMAIL_HOST")
+EMAIL_PORT = env.int("EMAIL_PORT")
+EMAIL_HOST_USER = env("EMAIL_HOST_USER")
+EMAIL_HOST_PASSWORD = env("EMAIL_HOST_PASSWORD")
 DEFAULT_EMAIL = "acceslibre@beta.gouv.fr"
 DEFAULT_FROM_EMAIL = f"L'équipe {SITE_NAME} <{DEFAULT_EMAIL}>"
 MANAGERS = [("Acceslibre", DEFAULT_EMAIL)]
```
Anto59290, line 268: I think we should also use this PR to remove https://pypi.org/project/dj-database-url/ and use something like `**env.db("DATABASE_URL")` to get the URL right

Anto59290: And same goes for cache I think

Anto59290 (commented):
Code changes are ok but I think we could use the new lib a bit more and remove another lib used only to configure the DB.

Anto59290 (approved):
Looks good

PR #608, opened 2023-06-12
Remove vendorized DSFR. First step.

mlvernay:
- Remove vendorized DSFR v1.0.0, now using packaged DSFR v1.4.0
- Make some app specific CSS adjustments
- Change lang selector, use a classic drop down from DSFR
- Remove hardcoded things in make messages command :facepalm: 
- Reduce landing images size, using the max size used in app
- Use 2 distincts template for radio button and checkboxes (was previously using boolradio.html)

TODO : 
- [x] fix street/satellite switch on leaflet
- [x] fix button radio
- [x] fix checkboxes


Anto59290 (commented):
Just one comment, we can talk about it if needed :thinking:

PR #614, opened 2023-06-15
Remove global map, useless with new dynamic map.

mlvernay:


Anto59290 (commented):
Looks good, I just don't get why they are so many changes in `locale/en/LC_MESSAGES/django.po`

mlvernay:
diff between compilemessages made by rosetta vs by django-admin

PR #630, opened 2023-06-20
New api for municipalities

Anto59290:
For QA:
- Test with multiple villages one postcode (Beuvry la foret)
- Test with one city multiple postcode (Paris, Lille)
- Test that the map refresh is still working in the parcours de contribution


static/js/api.js
```diff
@@ -89,50 +89,115 @@ function saveUserLocation(loc) {
...
+  clone["postcode"] = postcode;
+  clone["text"] = `${result["municipality"]} (${postcode})`;
+  return clone;
+}
 
-    return { q, results };
-  } catch (err) {
-    // most likely a temporary network issue
-    console.warn(`error while searching location: ${err}`);
+function createOneResultPerPostcode(api_results) {
```
Anto59290, line 153: @mlvernay This is where I took a decision that could be discussed. 

Now that we have an API that send multiple postcodes for one city we have 2 choices:

- Option 1: Send the list of postcodes to the backend and handle it there
- Option 2: Duplicate results and show one item in the auto-complete per postcode

I was going for option 1 when I met an issue. This means when looking for Paris it will look for all the postal codes (all arrondissements) at the same time, which makes the search pretty useless. (This is also true for smaller cities with multiples postcode like Lille but in this case I would consider it OK).



I went for option number 2 which adds a bit of JS here. I need more time to test, but from there the backend should do the job of filtering on the correct postcode. The main drawback of this is that the list in the autocomplete can be very long, see images

Capture d’écran de 2023-06-20 11-27-07

Capture d’écran de 2023-06-20 11-26-31

mlvernay: Go for option 1.

erp/views.py
```diff
@@ -208,28 +208,24 @@ def _cleaned_search_params_as_dict(get_parameters):
 
 def _clean_address(where):
     """
-    where is a string as returned by geoloc API on frontend side. It returns city and postal_code, this is pure string
+    where is a string as returned by geoloc API on frontend side. It returns city and code_departement, this is pure string
```
mlvernay, line 211: I would have use an english var name for code_departement.
Everything else is OK

mlvernay (approved):

erp/views.py
Anto59290: Yes I did not had any idea how to call this :thinking:

mlvernay: I would have naively use `department` but according to wordreference, it can be left untranslated

PR #682, opened 2023-07-10
Optimize API Key verification

Anto59290:
This commit optmize the time spent to verify an API key.

Using two classes in `DEFAULT_PERMISSION_CLASSES` was causing an increase in latency for all API views. This is due to the fact that both the `HasAPIKey` and `IsAllowedForAction` classes are checking for the existence of the key using `get_from_key` (this calls a `is_valid` method which is doing some criptographic verifications that are slow).

As the `IsAllowedForAction` class is already checking the key, we don't really need the `HasAPIKey` class.

This commit also add a custom span so that in the future we know what is happening in the traces (avoiding a big blank in sentry)

### DIFF
--- a/api/permissions.py
+++ b/api/permissions.py
@@ -1,3 +1,4 @@
+import sentry_sdk
 from django.conf import settings
 from rest_framework import permissions
 from rest_framework_api_key.models import APIKey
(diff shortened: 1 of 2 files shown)


api/permissions.py
```diff
@@ -9,9 +10,9 @@ class IsAllowedForAction(permissions.BasePermission):
     def has_permission(self, request, view):
         key = request.META["HTTP_AUTHORIZATION"].split()[1]
```
mlvernay, line 11: maybe use a .get here, to prevent from having a KeyError. Was previously used in combination with HasAPIKey which was checking this.

PR #692, opened 2023-07-11
Try to remove warnings on erp detail page

Anto59290:


erp/forms.py
```diff
@@ -339,48 +337,43 @@ class Meta:
...
+    nullable_values = [None, "", []]
+
+    def _build_data_for_field(self, field):
+        form_field = self[field["id"]]
+        field_value = form_field.value()
+        if field_value in self.nullable_values:
+            return
+        label, values = self.get_label_and_values(form_field.name, field_value, field["choices"], field["unit"])
+        return {
+            "name": form_field.name,  # TODO i am not sure we need the name
```
Anto59290, line 349: Todo check if this is really needed

erp/forms.py
```diff
@@ -339,48 +337,43 @@ class Meta:
...
+            "is_comment": form_field.field.widget.template_name.endswith("textarea.html"),
+        }
 
-    def get_accessibilite_data(self, flatten=False):
+    def get_accessibilite_data(self):
         data = {}
-        for section, section_info in self.fieldsets.items():
-            data[section] = {
+        for section_name, section_info in self.fieldsets.items():
+            section_data = {
```
Anto59290, line 359: It looks like section data is re-creating what is section_info, I believe we could skip this part but did not get it to work properly for now

erp/forms.py
```diff
@@ -339,48 +337,43 @@ class Meta:
...
                 "icon": section_info["icon"],
                 "tabid": section_info["tabid"],
                 "description": section_info["description"],
-                "edit_route": section_info["edit_route"],
-                "fields": [],
+                "edit_route": section_info["edit_route"],  # TODO : is this needed ?
```
Anto59290, line 363: Not sure if needed here

templates/erp/includes/fields.html
```diff
@@ -1,41 +1,31 @@
 {% load a4a %}
 {% load static %}
 {% load i18n %}
-{% if fields %}
-    <ul class="p-0 m-0">
-        {% for field in fields %}
-            {% render_field field.value as field_value %}
-            {% if field_value %}
```
Anto59290, line 8: Removed, the logic was added in the python to avoid:

- Building all the dicts
- Removing the one with all empty field (python)
- Hiding those without values (template)

All the work is now done in python to avoid back and forth to understand why a value is displayed or not

templates/erp/includes/fields.html
```diff
@@ -1,41 +1,31 @@
...
+    {% for field in fields %}
+        {% render_field field.value as field_value %}
+        <li class="a11y-field border-bottom">
+            <p class="h6 m-0 p-0 font-weight-normal">
+                {% if field.is_comment %}<i aria-hidden="true" class="icon icon-bubble"></i>{% endif %}
+                {{ field.label }}
+                {% if field.values and not field.is_comment %}
+                    :
+                    {% for value in field.values %}
+                        {% if value|startswith:"http" %}
```
Anto59290, line 14: I am not sure why this should be allowed @mlvernay maybe URLs in free text / comments ? But then why adding the link to the page, this looks like a weird behavior in this case

Anto59290: Can't be in free text because of `field.values and not field.is_comment`

erp/forms.py
Anto59290: (I was missing the deepcopy, which led to the page working once but not twice)

templates/erp/includes/fields.html
Anto59290: I added this in the new card I created in order to see what should be done in this case.

mlvernay (approved):

PR #702, opened 2023-07-13
Upgrade crispy forms

mlvernay:
I don't know why this seems now to pass the tests.


mlvernay:
Still the case on CI side

PR #722, opened 2023-07-20
Change layout, add component.

mlvernay:
Change layout: add a new row to display errors + activity suggestion
Add a new component to avoid the possibility to select an activity + make a
suggestion.


static/js/ui/ActivitySelect.js
```diff
@@ -0,0 +1,16 @@
+function ActivitySelect(root) {
+  root.addEventListener('change', function () {
+    let aNoActivity = document.getElementById('no_activity')
+    let spanActivitySuggested = document.getElementById('activity_suggested')
+    document.querySelector('#activity_suggested>span').innerHTML
```
Anto59290, line 5: I am not sure this is needed :thinking:

PR #736, opened 2023-07-25
Avoid creation of revision when no changes on ERP

Anto59290:
This commit will avoid the creation of a new revision on the ERP object, when it's edited without any changes in the first step of the contribution edit flow ("infos" step)

This is the first step to make sure that going thought the edit flow without any change does not trigger the creation of a revision (more work needs to be done in this direction).

- Add django-webtest to make sure it is easier to test forms / contribution flows
- Add more fixture to ease testing
- Make sure we call the .save() method only when some changes were actually made:
    - Use Decimal for the initial data, this will be used in the Django
internal to determine if the data has changed (`has_changed` method of the field object). As the field use the form is a DecimalField the cleaned data will be using a Decimal type. In order for the comparison to check for equality properly we need the types to be identical
    - Check for changes in the form using the `has_changed` method of
the form
- Add id in HTML to ease testing
- Add tests on the two cases we consider (edit an ERP without modifications, edit an ERP with modifications)

### DIFF
--- a/erp/views.py
+++ b/erp/views.py
@@ -2,6 +2,7 @@ import datetime
 import json
 import re
 import urllib
+from decimal import Decimal
 
 import waffle
 from django.conf import settings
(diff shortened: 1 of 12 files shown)


erp/views.py
```diff
@@ -864,27 +865,33 @@ def contrib_admin_infos(request):
...
-            erp = form.save()
-            if request.user.is_authenticated and erp.user is None:
-                erp.user = request.user
-                erp.save()
+            if form.has_changed():
+                erp = form.save()
+                if request.user.is_authenticated and erp.user is None:
+                    erp.user = request.user
+                    erp.save()
             if erp.activite.slug == "autre":
```
mlvernay, line 884: this can be under the `form.has_changed` block I guess, and so, we could have `if form.is_valid() and form.has_changed()`

Anto59290: I did what I could to take this into account, but we also need to redirect the user on the next step

mlvernay (approved):

PR #753, opened 2023-07-31
Filters and shortcuts V1.

Anto59290:


erp/views.py
```diff
@@ -256,6 +259,26 @@ def _get_or_create_api_key():
     return api_key
 
 
+def get_equipments_groups():
+    # TODO add icon
+    EquipmentGroup = namedtuple("EquipmentGroup", ["name", "slug", "equipements"])
```
Anto59290, line 264: @mlvernay Missing icon and suggestions here

erp/views.py
```diff
@@ -256,6 +259,26 @@ def _get_or_create_api_key():
...
+
+    # TODO example values, replace by correct values
+    for_pmr = EquipmentGroup(
```
Anto59290, line 267: This is purely for demonstration, I did not follow any logic or what is planned in the miro document

erp/views.py
```diff
@@ -275,6 +298,7 @@ def search(request):
         "map_api_key": _get_or_create_api_key(),
         "dynamic_map": True,
         "equipments": get_equipments() if waffle.flag_is_active(flag_name="EQUIPMENT_FILTERS", request=request) else [],
+        "equipments_group": get_equipments_groups(),  # TODO add flag
```
Anto59290, line 301: This needs to take into account the feature flag

templates/search/tag_with_label.html
```diff
@@ -0,0 +1,20 @@
...
+           class="hidden"
+           value="{{ slug }}"
+           {% if slug in params %}checked{% endif %}
+           autocomplete="off">
+    <label class="fr-label invisible ml-0" for="{{ slug }}"></label>
+    <button class="fr-tag a4a-label-tag mb-1"
+        aria-pressed="{% if slug in params %}true{% else %}false{% endif %}"
+        data-fr-js-toggle="true"
+        type="button"
+        data-event-type="{% if group %}groupLabelClicked{% else %}filterAdded{% endif %}"
```
Anto59290, line 14: This could / should be handled in another way to keep the HTML generic, maybe just pass the string of the expected event type

static/js/index.js
```diff
@@ -70,3 +70,26 @@ window.onload = function () {
...
+      inputFilter.checked = !inputFilter.checked
+      needRefresh = true
+    }
+  })
+
+  if (needRefresh) {
+    document.dispatchEvent(new Event('filterAdded'))
+  }
+
+  // TODO
```
Anto59290, line 93: Done, can be removed

Anto59290:
@mlvernay Not tested / done:

- Multiple filters
- Icons
- Suggestions of filters

erp/views.py
Anto59290: suggestions are done

PR #842, opened 2023-09-14
Rework contact page.

mlvernay:
TODO : 
- [ ] manage default topic, change link from erp page (to confirm)
- [ ] suppress "i am not a robot" when user is authenticated
- [ ] delete blue block on the right hand
- [ ] change text on yellow info block
- [ ] suppress sentence about *
- [ ] adjust colors to be DSFR compliant on info/alert boxes


locale/en/LC_MESSAGES/django.po
```diff
@@ -3834,6 +3807,31 @@ msgstr ""
 "\n"
 "If you don't have an API key yet, <a href=\"%(settings.SITE_ROOT_URL)s/contact/api_key\">contact us</a>, we have, at first, no reason to refuse :-)"
 
+#~ msgid "Nous ne pouvons pas vous fournir une assistance vis à vis des services ou prestations proposés."
```
Anto59290, line 3810: Maybe we can remove those ?

PR #854, opened 2023-09-19
Change logo of "André de CABO" partner

Anto59290:


mlvernay:
Height max 120px would avoid a on the fly resizing. I will manage this

mlvernay (approved):

PR #860, opened 2023-09-25
New implementation of widget referer

Anto59290:


mlvernay (approved):
Maybe add a test with views == 2 and keeping nb WidgetEvent == 1

PR #863, opened 2023-09-25
Activity picker

Anto59290:


mlvernay:
When clicking "I can't find the activity", it should: 
- set "Other" as activity in the autocomplete
- hide the autocomplete
- show the input text #new_activity

mlvernay (approved):

PR #905, opened 2023-10-09
Make the API doc public even with no token

Anto59290:
Allow any user to see the documentation of the API even without a token / API Key.

- Make sure the schema is available for everyone (with the permission class)
- Make sure the schema also display the routes which needs auth / permissions
- Fix some code, when building the schema DRF will call the view without a request (request is None in `_get_paths_and_endpoints` and `create_view`)


mlvernay:
- [ ] put the OK button on same line as the input text (initial Sophie's request)
- [ ] put a text decoration underline on the contact link (initial Sophie's request)
- [ ] when trying out the api, the message "for internal uses only" is a bit weird in this context

PR #908, opened 2023-10-10
Ignore shift and control on autocomplete, add a required on what.

mlvernay:
- Ignore shift and control, those keys should not empty the search values, i.e. if you perform a shift tab to go to the activity field, it should not wipe the where autocomplete
- add a required on the what field, so that it acts as the activity field, with a direct message if empty.

### DIFF
--- a/static/js/ui/SearchWhere.js
+++ b/static/js/ui/SearchWhere.js
@@ -128,7 +128,8 @@ function SearchWhere(root) {
   // Invalidate lat/lon on every key stroke in the search input, except when user tabs
   // out of the field or selects and entry by pressing the Enter key.
   autocomplete.input.addEventListener('keydown', (event) => {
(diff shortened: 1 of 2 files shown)


static/js/ui/SearchWhere.js
```diff
@@ -128,7 +128,8 @@ function SearchWhere(root) {
   // Invalidate lat/lon on every key stroke in the search input, except when user tabs
   // out of the field or selects and entry by pressing the Enter key.
   autocomplete.input.addEventListener('keydown', (event) => {
-    if (event.key != 'Tab' && event.key != 'Enter') {
+    var ignoringKeys = ['Tab', 'Enter', 'Shift', 'Control']
```
Anto59290, line 131: `ignoredKeys` sounds more appropriate to me in this case

Anto59290 (approved):

PR #937, opened 2023-10-17
SP importer, make it more robust

mlvernay:


Anto59290 (approved):
I believe we could review this using a serializer

PR #954, opened 2023-10-23
 Working on using new activity field 

Anto59290:


mlvernay:
Missing a prevent default on "i can not find activity" in contrib. Non critical

mlvernay (approved):

PR #963, opened 2023-10-24
WIP - Update to parcel 2

Anto59290:
Update the project to parcel 2 instead of parcel-bundler (parcel 1).

Issues and documentation:

- DSFR Css is not valid : https://github.com/GouvernementFR/dsfr/issues/576 (thanks @rik for the issue)
- Fix images paths : https://github.com/parcel-bundler/parcel/issues/7726

Todo:

- [x] Check if there is a better way to handle paths to statics (this works but should we be using this parameter ?)
- [x] Check if we can remove core-js and why babel-preset-env / babel-core are in the "prod" dependencies instead of the dev ones ?
- [x] Should parcel be only in the dev dependencies ?
- [x] Remove parcel-entry.js (no still needed)
- [x] Check why files are now called parcel-entry.css and .js


mlvernay:
Everything seems ok with the last changes I push.

PR #967, opened 2023-10-25
browser geoloc

mlvernay:
- Increase the timeout: on FF it was reaching the timeout before being able to retrieve the coords. It is much more slower on FF than on Chrome
- Reduce code complexity, was here only to be able to display a label


mlvernay:
It breaks the search :warning:

Anto59290 (approved):

PR #1021, opened 2023-11-13
Add command to remove duplicates from Acceo

Anto59290:
- Factorize with existing command
- Adapt search of duplicates

### DIFF
--- a/erp/management/commands/base_duplication.py
+++ b/erp/management/commands/base_duplication.py
@@ -1,0 +1,69 @@
+from django.core.management.base import BaseCommand
+
+from erp.exceptions import MergeException
+from erp.models import Erp
+
+
+class BaseDuplicationCommand(BaseCommand):
+    to_delete = 0
+    unhandled = 0
+    help = "Best effort to remove all the duplicates"
+
+    def add_arguments(self, parser) -> None:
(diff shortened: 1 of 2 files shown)


erp/management/commands/base_duplication.py
```diff
@@ -0,0 +1,69 @@
+from django.core.management.base import BaseCommand
+
+from erp.exceptions import MergeException
+from erp.models import Erp
+
+
+class BaseDuplicationCommand(BaseCommand):
```
mlvernay, line 7: It looks like we will create duplicates :-D 
I would have named it Deduplication or RemoveDuplicate, something like this.

erp/management/commands/base_duplication.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+
+    def _get_duplicates(self, erp):
+        raise NotImplementedError
+
+    def _get_queryset(self):
+        raise NotImplementedError
+
+    def _delete_duplicates(self, duplicates):
+        self.to_delete += len(duplicates)
```
mlvernay, line 26: could/should be duplicates.count(), no?

Anto59290:
Merged done via another PR

PR #1025, opened 2023-11-13
Adapt command to remove duplicates for town halls

Anto59290:
This change will lead to the number of "unhandled" cases to be falsy in dry mode because all the ERPs will be counted (see test).

As we are no longer sure that we keep the ERP coming from service public we need to copy the ASP Id when it exists.

### DIFF
--- a/erp/management/commands/remove_duplicate_mairie.py
+++ b/erp/management/commands/remove_duplicate_mairie.py
@@ -1,7 +1,7 @@
 from django.core.management.base import BaseCommand
 from geopy import distance
 
-from erp.exceptions import MergeException
+from erp.exceptions import MergeException, MultipleAspIdForDuplicates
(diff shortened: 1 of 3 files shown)


erp/management/commands/remove_duplicate_mairie.py
```diff
@@ -57,6 +57,13 @@ def _merge_and_delete(self, erp, duplicates):
             duplicates[0].delete()
         self.to_delete += 1
 
+    def _get_asp_id(self, duplicates):
+        ids = set([e.asp_id for e in duplicates if e.asp_id])
+        if len(ids) == 1:
+            return list(ids)[0]
```
mlvernay, line 63:
```suggestion
            return ids.pop()
```

mlvernay (approved):

PR #1047, opened 2023-11-20
Add script to create and update municipalities

Anto59290:
Using the JSON file in order to make the process faster than using the API. The JSON file also contains the links between old municipalities and new ones (which is not the case for the API).

- Write a command that will create new municipalities. Note that I had to get the geometry from the API to be able to create the municipality.
- Update the existing municipalities (without touching geometry and contour, this will be done in a separate script).
- Remap existing ERPs to new municipalities when possible


erp/dataclasses.py
```diff
@@ -0,0 +1,24 @@
...
+    def from_json(cls, json):
+        return cls(
+            type_of=json["type"],
+            nom=json["nom"],
+            code_insee=json["code"],
+            departement=json["departement"],
+            code_postaux=json.get("codesPostaux"),
+            population=json.get("population"),
+            parent_municipality=json.get("chefLieu"),
+        )
```
mlvernay, line 24: We could have use english everywhere in this dataclass no ?

erp/management/commands/update_municipalities.py
```diff
@@ -0,0 +1,78 @@
...
+        except Commune.DoesNotExist:
+            return
+
+        commune.obsolete = True
+
+        if new_insee:
+            erps = Erp.objects.filter(commune_ext=commune)
+            if erps:
+                new_commune = Commune.objects.get(code_insee=new_insee)
+                erps.update(commune_ext=new_commune)
```
mlvernay, line 48: We should update only if --write, no?

erp/management/commands/update_municipalities.py
```diff
@@ -0,0 +1,78 @@
...
+        if self.write:
+            commune.save()
+
+    def handle(self, *args, **options):
+        self.write = options["write"]
+        data = json.load(open("communes.json"))
```
mlvernay, line 54: where does this come from ?

Anto59290: It's written in the file ;)

PR #1056, opened 2023-11-22
Check business_status with Outscraper

mlvernay:


erp/management/commands/outscraper_clean_closed.py
```diff
@@ -0,0 +1,62 @@
...
+        else:
+            print("Dry run mode, no DB action, use --write to apply this deletion")
+
+    def handle(self, *args, **options):
+        self.write = options["write"]
+        self.start_pk = options.get("start_pk")
+
+        client = ApiClient(api_key=settings.OUTSCRAPER_API_KEY)
+
+        two_month_ago = timezone.now() - timedelta(days=56)
```
Anto59290, line 43: Why 56 ?

Anto59290 (commented):
Maybe we should add a test on it too

erp/management/commands/outscraper_clean_closed.py
```diff
@@ -0,0 +1,69 @@
...
+        else:
+            print("Dry run mode, no DB action, use --write to apply this deletion")
+
+    def handle(self, *args, **options):
+        self.write = options["write"]
+        self.start_pk = options.get("start_pk")
+
+        client = ApiClient(api_key=settings.OUTSCRAPER_API_KEY)
+
+        two_month_ago = timezone.now() - timedelta(days=options["nb_days"])
```
Anto59290, line 51: With the new parameter the variable has an improper naming I believe (proposals: date_limit, date_nb_days_ago

tests/erp/test_commands.py
```diff
@@ -356,3 +355,59 @@ def test_max_results(self, mocker):
...
+        ]
+    ]
+
+    @pytest.mark.django_db
+    def test_initial(self, mocker):
+        ErpFactory(nom="Le lard", commune="Lyon", voie="Rue Trouvier", numero=173)
+        mocker.patch("outscraper.ApiClient.google_maps_search", return_value=self.initial_outscraper_response)
+
+        call_command("outscraper_clean_closed", nb_days=0, write=False)
+        assert Erp.objects.filter(nom="Le lard").count() == 1, "should have deleted it"
```
Anto59290, line 400: We don't want to delete in this case (right ?) so the assert message should be the other way around I believe ("should not have deleted the object")

Anto59290 (approved):

PR #1079, opened 2023-12-04
POC Scrapfly

mlvernay:


erp/management/commands/scrapfly_acquisition.py
```diff
@@ -0,0 +1,266 @@
...
+            "Lavabo bas adapté aux personnes à mobilité réduite",
+            "Toilettes surélevées",
+            "Aides visuelles : braille",
+            "Étages supérieurs accessibles par ascenseur",
+            "Étages supérieurs accessibles uniquement par les escaliers",
+            "Aides visuelles : panneaux tactiles",
+        ):
+            if key_to_ignore not in access_copy:
+                continue
+            access_copy.remove(key_to_ignore)
```
Anto59290, line 80:
```suggestion
            access_copy.pop(key_to_ignore, None)
```
I think this could work and would be simpler, what do you think ?

erp/management/commands/scrapfly_acquisition.py
```diff
@@ -0,0 +1,266 @@
...
+
+        if access_copy:
+            print(f"Extra access keys: {access_copy}")
+        return accessibility
+
+    def _parse_search_page(self, result):
+        print("parsing search page: {}", result.context["url"])
+        hotel_previews = []
+        for hotel_box in result.selector.xpath('//div[@data-testid="property-card"]'):
```
Anto59290, line 89: Maybe the selectors could be constant at the start of the file for better readability and maintenance ?

erp/management/commands/scrapfly_acquisition.py
```diff
@@ -0,0 +1,266 @@
...
+                and "non_field_errors" in e.get_codes()
+                and "duplicate" in e.get_codes()["non_field_errors"]
+            ):
+                erp_duplicated = Erp.objects.get(pk=int(e.detail["non_field_errors"][1]))
+                return self._create_or_update_erp(result, existing_erp=erp_duplicated)
+            return None
+
+        action = "CREATED"
+        if existing_erp:
+            action = "UPDATED"
```
Anto59290, line 178:
```suggestion
        action = "UPDATED" if existing_erp else "CREATED"
```

Anto59290 (commented):
Added a few comments here and there

erp/management/commands/scrapfly_acquisition.py
mlvernay: access is a list, not a dict

PR #1081, opened 2023-12-04
Migrate contact email (to admins and sender) to Brevo.

mlvernay:


contact/models.py
```diff
@@ -24,7 +23,6 @@ class Message(models.Model):
         (TOPIC_CONTACT, translate("Prise de contact avec Acceslibre")),
         (TOPIC_API, translate("API")),
         (TOPIC_SIGNALEMENT, translate("Signaler une malveillance")),
-        (TOPIC_VACCINATION, translate("Vaccination")),
```
Anto59290, line 27: What will happen to existing message with this topic ?

mlvernay: I had the same question, but I think we don't care at all about those stored message. I discover the Message model while doing this PR :-D

PR #1123, opened 2024-01-02
Always sync user to Brevo

mlvernay:
When following the nominal sign up funnel, while creating an ERP, the ERP is first created with no user, then the user is created and associated to the ERP. In that case we were not synch'ing user to Brevo as we were only synch'ing user when ERP is created.


mlvernay:
This PR needs to be reworked, nb_erp_created need to be fixed as well.

mlvernay:
TODO add some tests for both cases
- [ ] creation with logged user
- [ ] creation with no user, registered at the end of the funnel
- [ ] update

PR #1175, opened 2024-01-31
Consider same name in 75m around as a duplicate in import.

mlvernay:
New definition of duplicates in import: now also searching by exact name in the 75m around.

### DIFF
--- a/erp/imports/serializers.py
+++ b/erp/imports/serializers.py
@@ -195,7 +195,7 @@ class ErpImportSerializer(serializers.ModelSerializer):
                     continue
 
                 if obj.get("latitude") is not None and obj.get("longitude") is not None:
-                    self._geom = Point((obj["latitude"], obj["longitude"]))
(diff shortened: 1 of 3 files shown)


mlvernay:
TODO : add a test

erp/imports/serializers.py
```diff
@@ -208,14 +208,27 @@ def validate(self, obj):
 
         existing = Erp.objects.find_duplicate(
             numero=obj.get("numero"),
-            voie=obj.get("voie"),
-            lieu_dit=obj.get("lieu_dit"),
             commune=obj["commune"],
             activite=obj["activite"],
+            voie=obj.get("voie"),
```
Anto59290, line 213: We could revert change if not mandatory

Anto59290: (not mandatory)

mlvernay: it was more to match the function signature.

PR #1216, opened 2024-02-20
Support audiodesc questions in convert from tally script.

mlvernay:


mlvernay:
Waiting for QA on cinema import

PR #1218, opened 2024-02-22
Launch acquisition every month.

mlvernay:


mlvernay:
@Anto59290 I'm really interested in having your though on this, it might have a smarter solution to do this

Anto59290:
@mlvernay  I had a look at it and it seems reasonable if we don't have any other solutions. Maybe some others things that could be done:
- Run a cron everyday that scrap 4 departement (Jan 1st scraps Ain, Aine, Allier, Alpes Haute-Provence, Jan 2nd scraps Haute-Alpes and so on). I am not sure if this is doable or not (maybe the scrap time for 4 departements is higher than the limit time of one off dyno ?)
- Trigger a celery tasks per departement / query tuple on the first day of the month and just let them do their job ? (In a separate queue not to mess with the rest ?)

PR #1230, opened 2024-02-26
Use dsfr-chart in stats page.

mlvernay:


mlvernay:
Increasing the built index.js. Too many dependencies. Keep it as it is.

PR #1232, opened 2024-02-27
Remove nearest ERPs on ERP detail page

Anto59290:
- Remove nearest ERPs at the bottom of the ERP page and in the header of the map
- Make sure we cannot move/drag/zoom the map on the ERP details page

This is part of the new design of the ERP detail page.


mlvernay (approved):
We will, then, have to drop the switch once deployed (`USE_GEOSPATIAL_SEARCH_IN_DETAIL`)

PR #1259, opened 2024-03-12
Generate new schema

Anto59290:
Try to update the schema.json to show the real state of the schema. A lot of things should be done here:
- No source of truth for Activities (prod is the truth). We will never be up-to-date on this. This also means we have a different result on our host (local db) and in the tests (test fixtures).
- The generation of the schema is manual, we have no proof the schema will be updated in the future (and copied to the other repo).
- The comparison of files makes it hard to debug when something goes wrong.

- Unskip the test
- Generate the new schema
- Try to make the schema valid by changing the type of examples (actual version is valid per https://www.itb.ec.europa.eu/json/tableschema/upload)


erp/export/static/schema.json
```diff
@@ -1586,9 +1627,16 @@
...
+          "bim",
+          "bmp",
+          "lsf",
+          "lpc",
+          "sts",
+          "autres"
+        ]
       },
-      "example": "Valeurs possibles: bim -> boucle à induction magnétique fixe, bmp -> boucle à induction magnétique portative, lsf -> langue des signes française (LSF), lpc -> langue française parlée complétée (LFPC), sts -> sous-titrage ou transcription simultanée, autres -> autres"
+      "example": "['bim', 'bmp', 'lsf', 'lpc', 'sts', 'autres']"
```
mlvernay, line 1639: This one might be hard to understand.

mlvernay (commented):
We could also refacto the `field.get("example", generate...) or generate` we saw together.

mlvernay (approved):

PR #1275, opened 2024-03-19
Allow to mark an ERP as up-to-date

Anto59290:
Allow any user to mark the data we show on the ERP details page as "up to date" in order to know when the data was last checked.


mlvernay:
I am OK on a pure tech point of view but I have some doubts on a product point of view. It seems very very easy to confirm any ERP as up to date. No need to be the ERP user (creator or attributed) ?

mlvernay:
Checked on a product point of view :heavy_check_mark: , let's go :rocket: !

PR #1276, opened 2024-03-20
Challenge refacto, new scoring method.

mlvernay:


stats/queries.py
```diff
@@ -46,18 +52,48 @@ def get_top_contributors():
...
-        erp__created_at__lt=stop_date,
-    )
-    challengers = get_user_model().objects.filter(email__in=emails_players_list)
-    top_contribs = (
-        challengers.annotate(erp_count_published=Count("erp", filter=filters, distinct=True))
-        .filter(erp_count_published__gt=0)
-        .order_by("-erp_count_published")
+def _get_nb_filled_in_info(access_fields):
+    fields_to_count = set(schema.get_a11y_fields()) - set(schema.get_free_text_fields())
+    nb = 0
```
Anto59290, line 57: Maybe something like this could be easier (not tested)
```suggestion
    values = [access_fields.get(f) for f in fields_to_count if access_fields.get(f) not in [None, "", []]]
    return len(values)
```
Or like this (improved readability ?)
```suggestion
    values = [access_fields.get(f) for f in fields_to_count]
    return len([v for v in values if v not in [None, "", []]])

```
Did I miss something ?

templates/challenge/detail_v2.html
```diff
@@ -0,0 +1,121 @@
...
+                        <small class="text-muted">
+                            {{ top_contribs.0.nb_access_info_changed }}
+                            {% if top_contribs.0.nb_access_info_changed > 1 %}
+                                {% translate "informations d'accessibilité renseignées" %}
+                            {% else %}
+                                {% translate "information d'accessibilité renseignée" %}
+                            {% endif %}
+                        </small>
+                    {% else %}
+                        ???
```
Anto59290, line 70: Left on purpose ?

mlvernay: image

stats/models.py
```diff
@@ -73,25 +73,22 @@ def get_absolute_url(self):
...
-                get_user_model().objects.filter(email__contains="@ville-antony.fr").values_list("email", flat=True)
-            )
-        else:
-            emails_players_list = self.players.all().values_list("email", flat=True)
-
-        classement, self.nb_erp_total_added = get_count_challenge(self.start_date, self.end_date, emails_players_list)
+        players_ids = list(set(self.players.all().values_list("id", flat=True)))
+
+        scores_per_user_id = get_challenge_scores(self.start_date, self.end_date, players_ids)
+        users = get_user_model().objects.filter(pk__in=[user_id for user_id, _ in scores_per_user_id]).only("username")
```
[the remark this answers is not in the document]
mlvernay: At this point scores_per_user_id is a list of tuples. 

```
[(2730, 1), (9644, 0)]
```

But we can use players_ids, I guess it is acceptable.

Anto59290 (approved):

PR #1289, opened 2024-03-25
Introduce new display of access data on ERP detail

Anto59290:
This commit adds 2cff9f71b6c9577589d07effe907478d76dbae11 and 092e292bfce43b8b06e563b8f760aa6190fd17de back to the branch. Also add new texts in the schema and use them in template tags.


mlvernay (approved):
Could have been more concise, but I guess we do not have to worry about that as it is temp code

PR #1318, opened 2024-04-08
Add new rule for "updated at" date on ERP page

Anto59290:
- Add new logic to take the most recent date between 3 dates
- Review the design of the display

### DIFF
--- a/erp/models.py
+++ b/erp/models.py
@@ -1013,6 +1013,12 @@ class Erp(models.Model):
         if needs_save:
             access_destination.save()
 
+    @property
+    def displayed_last_updated_date(self):
+        dates = [d for d in [self.checked_up_to_date_at, self.created_at, self.updated_at] if d]
+        if dates:
+            return max(dates)
+
 
 @reversion.register(
     ignore_duplicates=True,
--- a/templates/erp/index.html
+++ b/templates/erp/index.html
@@ -41,11 +41,13 @@
                 {% include "common/map.html" with map_classes="a4a-app-erp-map rounded shadow-sm" %}
             </aside>
             <div class="d-flex mb-2">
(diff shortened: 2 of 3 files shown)


templates/erp/index.html
```diff
@@ -41,11 +41,13 @@
                 {% include "common/map.html" with map_classes="a4a-app-erp-map rounded shadow-sm" %}
             </aside>
             <div class="d-flex mb-2">
-                <h5 class="h5 inline mr-2">{% translate "Informations sur l'accessibilité" %}</h5>
-                {% if erp.checked_up_to_date_at %}
-                    {% blocktranslate trimmed with date=erp.checked_up_to_date_at|date %}
-                        - mises à jour le {{ date }}
-                    {% endblocktranslate %}
+                <h5 class="h5 inline bold mr-2">{% translate "Informations sur l'accessibilité" %}</h5>
```
mlvernay, line 44: Maybe remove the h5 class as you removed the associated css style, no?

erp/models.py
```diff
@@ -1013,6 +1013,10 @@ def merge_accessibility_with(self, erp, fields=None):
         if needs_save:
             access_destination.save()
 
+    @property
+    def displayed_last_updated_date(self):
+        return max([d for d in [self.checked_up_to_date_at, self.created_at, self.updated_at] if d])
```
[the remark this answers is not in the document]
Anto59290: This could not really happens I think (except on a model that is not yet saved ?) I'll make it safer anyway

templates/erp/index.html
Anto59290: The h5 class is still needed in this case because it triggers more styles at other places in the CSS (for example the font size)

mlvernay (approved):

PR #1319, opened 2024-04-09
Add labels to new ERP page

Anto59290:
Add labels (Tourisme et Handicap, Handiplage, etc.) to the new ERP design page.


mlvernay (approved):
OK for me, even if I would have resized the image to fit our usage and reduce the load.

PR #1323, opened 2024-04-11
Add filters on ERP detail page

Anto59290:
Add the possibility for the user to filter on one or multiple types of handicap. By doing so we only display the data we think is appropriate for this type of handicap.

- Re-use tag_with_label.html which was not used in the code
- Implement filters with icons and JavaScript logic
- Add the type of handicap on all the data of the page (all the <li> tags)

### DIFF
--- a/templates/erp/includes/filters.html
+++ b/templates/erp/includes/filters.html
@@ -1,0 +1,11 @@
+{% load i18n %}
+<div class="d-flex row p-2">
+    {% translate "Difficulté à voir" as visual_text %}
+    {% translate "Difficulté à marcher" as walking_text %}
+    {% translate "Difficulté à entendre" as deaf_text %}
+    {% translate "Difficulté à comprendre" as understand_text %}
+    {% include "search/tag_with_label.html" with slug="visual" input_name="erp_filter" eventType="filterClicked" name=visual_text icon="diff-vision" %}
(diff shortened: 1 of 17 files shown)


Anto59290:
TODO:

- [ ] Check modification on JS file
- [ ] Test keyboard and voice over accessibility (same code as the search filter so it should be ok)

templates/erp/includes/filters.html
```diff
@@ -0,0 +1,6 @@
+<div class="d-flex row p-2">
+    {% include "search/tag_with_label.html" with slug="visual" input_name="erp_filter" eventType="filterClicked" name="Difficulté à voir" icon="diff-vision" %}
+    {% include "search/tag_with_label.html" with slug="walking" input_name="erp_filter" eventType="filterClicked" name="Difficulté à marcher" icon="diff-walking" %}
+    {% include "search/tag_with_label.html" with slug="hearing" input_name="erp_filter" eventType="filterClicked" name="Difficulté à entendre" icon="deaf" %}
+    {% include "search/tag_with_label.html" with slug="understand" input_name="erp_filter" eventType="filterClicked" name="Difficulté à comprendre" icon="diff-understand" %}
+</div>
```
mlvernay, line 6: Are the names translated later on? Looks like no
We might have to create vars at this point.

PR #1345, opened 2024-04-16
Show verbose text for dispositif appel type

Anto59290:
Show the verbose text instead of the database value for `entree_dispositif_appel_type`. This makes the display more explicit.

### DIFF
--- a/erp/models.py
+++ b/erp/models.py
@@ -1709,9 +1709,19 @@ class Accessibilite(models.Model):
         )
 
     def get_labels_display(self):
+        if not self.labels:
+            return
         label_to_text = {k: str(v) for k, v in schema.LABEL_CHOICES}
         return [(label, label_to_text.get(label)) for label in self.labels]
 
     def get_accueil_audiodescription_text(self):
+        if not self.accueil_audiodescription:
+            return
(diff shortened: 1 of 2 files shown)


erp/models.py
```diff
@@ -1715,3 +1715,7 @@ def get_labels_display(self):
     def get_accueil_audiodescription_text(self):
         equipment_to_text = {k: str(v) for k, v in schema.AUDIODESCRIPTION_CHOICES}
         return ",".join([equipment_to_text.get(equipment) for equipment in self.accueil_audiodescription])
+
+    def get_entree_dispositif_appel_type_text(self):
+        equipment_to_text = {k: str(v) for k, v in schema.DISPOSITIFS_APPEL_CHOICES}
+        return [equipment_to_text.get(equipment) for equipment in self.entree_dispositif_appel_type]
```
mlvernay, line 1721:
```suggestion
        return [equipment_to_text.get(equipment) for equipment in (self.entree_dispositif_appel_type or []])
```
Or an early return. 

We have a ~23k ERPs with `None` as value, which would raises a TypeError (NoneType not iterable).

Same for above helpers.

mlvernay (approved):

mlvernay (approved):

PR #1393, opened 2024-05-06
Fix display of negative text for ramp.

mlvernay:


templates/erp/includes/access_outside_path.html
```diff
@@ -27,7 +27,7 @@ <h4 class="pt-2 h4" data-filter-title="true">
                     {{ sentence }} {{ rampe }}
                 {% endblocktranslate %}
             </li>
-        {% elif access.cheminement_ext_rampe is False %}
+        {% elif access.cheminement_ext_rampe == "aucune" %}
```
Anto59290, line 30: By reading the definition of has_ramp_exterior_path I think this could be `elif acces.cheminement_ext_rampe`but I am not 100% sure

mlvernay: I am OK, but more readable like this, no?

mlvernay: But without  `== "aucune"` it would display the sentence in case of None, no? That's not what we want (I guess)

PR #1444, opened 2024-05-22
Fix postcodes

Anto59290:


Anto59290:
To test with a commune that has multiple postcode and is not a prefecture (-VILLENEUVE D'ASCQ (59493, 59494, 59650)) / -AIX-EN-PROVENCE (13080, 13090, 13100) / etc

PR #1463, opened 2024-05-27
Hide empty section on ERP page for old browsers

Anto59290:
The `has(+ ul:empty)` CSS rule is not handled properly on old browsers (including those used in the french administration) so we are adding this JavaScript logic to hide the title when needed. I am also leaving the CSS rule is anyone don't want to use the JavaScript but has a decent browser.

### DIFF
--- a/static/js/ui/FilterData.js
+++ b/static/js/ui/FilterData.js
@@ -8,18 +8,18 @@ function hasVisibleChildren(element) {
   return visibleChildren
 }
 
-function filterData(root) {
+function listenToFilterClicked(root) {
   const dataToToogle = root.querySelectorAll('[data-filters]')
   document.addEventListener('filterClicked', () => {
     const inputFilters = document.querySelectorAll('[name=erp_filter]:checked')
     const activeFilters = Array.from(inputFilters).map((filter) => filter.value)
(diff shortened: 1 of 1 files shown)


static/js/ui/FilterData.js
```diff
@@ -36,15 +36,25 @@ function filterData(root) {
...
 
-    titlesToToogle.forEach((titleElement) => {
-      const shouldBeVisible = hasVisibleChildren(titleElement.nextElementSibling)
-      if (shouldBeVisible === true) {
-        titleElement.classList.remove('hidden')
-      } else {
-        titleElement.classList.add('hidden')
-      }
-    })
+    hiddenEmptyTitles(root)
```
mlvernay, line 39: I would have name it `toggleEmptyTitle` (or `hideEmptyTitles`)
Not a big deal

static/js/ui/FilterData.js
```diff
@@ -36,15 +36,25 @@ function filterData(root) {
...
   })
 }
 
+function hiddenEmptyTitles(root) {
+  const titlesToToogle = root.querySelectorAll('[data-filter-title]')
```
mlvernay, line 44: `titlesToToggle`

Toogle does not exist AFAIK

mlvernay (approved):

PR #1478, opened 2024-05-31
Associate appropriate slug for wheelchair

mlvernay:
Was causing a duplicate id in form


Anto59290:
I don't think we can merge this as is, I believe the specs that needs to be correct before and after is that for this page Difficulty walking and wheelchair are equivalent filters (they should show the same data). This is true on master, and not true on this branch.

PR #1552, opened 2024-07-01
Stats from other matomo

Anto59290:
Needs https://github.com/MTES-MCT/acceslibre/pull/1551 to be merged first.

Also need to set 2 env var correctly on recette and prod


Anto59290:
I did the change on recette to make sur we can QA this

mlvernay (approved):

PR #1561, opened 2024-07-08
Allow to revert changes made by one user

Anto59290:
Add a command to revert the changes made by one user if they are still the actual value.


mlvernay (commented):
I would have use `version.revision.revert()`
Probably better to talk about it orally

PR #1564, opened 2024-07-11
Send async email in import

mlvernay:


mlvernay:
Does not seem to work :thinking:

PR #1576, opened 2024-07-22
Allow to search ERP inside a departement

Anto59290:
Allow to filter the location by departement name or number. Make sure we don't refresh the map on load in this case (per specs).


mlvernay:
Does not seem to work with dpt 37 (we have ERPs there), it returns an empty map with a world zoom out.
http://127.0.0.1:8000/recherche/?where=Indre-et-Loire+%2837%29&code=37&search_type=departement

core/settings.py
```diff
@@ -412,6 +412,7 @@ def expose_site_context(request):
 ADRESSE_DATA_GOUV_SEARCH_TYPE_HOUSENUMBER = "housenumber"
 ADRESSE_DATA_GOUV_SEARCH_TYPE_STREET = "street"
 IN_MUNICIPALITY_SEARCH_TYPE = "in_municipality"
+IN_DEPARTEMENT_SEARCH_TYPE = "departement"
```
mlvernay, line 415: Inconsistent naming between this line and the above one.
(`in_` prefix + FR/EN)

mlvernay (approved):

PR #1578, opened 2024-07-23
Migrate the deletion of global stats

Anto59290:


Anto59290:
@mlvernay Something else came with

mlvernay:
OK for the thing which came with :)
The only diff with 0022 is the verbose name, I guess it has been added after the 0022.

mlvernay (approved):

PR #2065, opened 2025-05-28
Remove no_activity from contrib start page

mlvernay:
We cannot drop more code, as no_activity is still in use in administrative page


Destinyyyy (approved):
🥳

PR #2227, opened 2025-09-17
[Frontend] Widget edge case aria-haspopup fix

Destinyyyy:
# Description
Some partners have issues integrating the widget because a query selector is too wide and selecting THEIR components instead of ours.
The solution is to add a differentiator such as data-owner="acceslibre" and add this to the selector.
Careful to the retrocompatibility, old scripts still need to work on the partner's end

## Trello ticket
- https://trello.com/c/c1PJdnnk/2361-pb-li%C3%A9-%C3%A0-laccess-widget


mlvernay:
QA OK

